PR #195, opened 2022-03-09
Usagov 78 new relic

drupalbee:
Fixes issue(s) # 78
https://cm-jira.usa.gov/browse/USAGOV-78

I updated the  Dockerfile-cms, docker-compose file to install and configure new relic agent. The license key and app name can be modified in the docker-compose.yml file. I also added and configured the New Relic module, it's app key can be modified from here admin/config/development/new-relic as it is using my free trial account. 
-
-


ednark (changes requested):
We will have to talk about how keys need to be passed before I can merge. I have some changed incoming.

PR #227, opened 2022-03-25
usagov-159-sp-footer-sm-margin-update

sinchagrenchik:
Fixes issue(s) # .

-
-
-


web/themes/custom/usagov/sass/_uswds-theme-custom-styles.scss
```diff
@@ -442,7 +442,7 @@ This style ensures that we can remove the padding-top from text in a card body.
 
 .grid-row.grid-gap-1 {
    @media (min-width: 31em){
-      margin-right: 1.5rem;
+      margin-right: 5rem;
    }
```
cwacht: We should not be changing the margin of grid-gap-1. It is a utility class which is meant to create a specific amount of margin.

After inspecting the website, it appears that these styles are being used to adjust the margin of the social links in the footer. We can use the `usa-footer__social-links` class for this instead.

We should also not be changing margin on `grid-row.grid-gap>*` on line 430 or `margin-bottom-neg-1` on line 434

cwacht (changes requested):

web/themes/custom/usagov/sass/_uswds-theme-custom-styles.scss
sinchagrenchik: I added the class to format just Spanish sm icons.  Please take a look.

cwacht: We really need to completely remove the grid-gap styles from the custom-styles file. I could do it as a new ticket, but it will effect the footer social media, so it makes sense to do it in this pull request.

sinchagrenchik: Thanks, Chris. Let me update the css.

cwacht (approved):

PR #235, opened 2022-03-30
USAGOV-130-Lang-Toggle-Reference

cwacht:
Fixes issue(s) # USAGOV-130.

- Create a field for content editors to select a page that the current page's language toggle will link to.


cwacht:
2 things:
* I could not determine how to exclude pages that have the same language of the current page.
* I had named the field `language_toggle_override`, but it should probably be `language_toggle` with our new plan to handle languages on the site.

cwacht:
I renamed the field to `language_toggle` and it is working, but:
* It still allows users to select a page in any language
* It still shows the title of the referenced page in the current page's language

Some parts of the template will need to be reworked after removing the Translation module

ednark (approved):

PR #280, opened 2022-04-23
usagov-212-autologout

skstarne:
Adds Autologout module


kyle-jester (approved):
All looks good to me.

PR #364, opened 2022-07-05
Update to "Federal Agencies" view

akf:
This is a small update to the "Federal Agencies" view I set up, adding alphabetic navigation. 
It assumes you have a "Directory Record" that starts with "A" so it can default to that. 

Note the addition of the view-views-summary.html.twig file to style the letters. 

I'm not convinced that this will be the right way to go.


thetypicalprogrammer (approved):
LGTM

PR #386, opened 2022-07-25
USAGOV-30 - State Directory Content Type

thetypicalprogrammer:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-30

## Checklist for the Developer

- [x] A link to the JIRA ticket has been added above.
- [ ] The JIRA ticket identifies the desired result of this change.
- [ ] The JIRA ticket contains clear acceptance criteria.
- [ ] The JIRA ticket contains clear testing steps.
- [ ] The JIRA ticket contains a description of "Done".
- [ ] Any preparation/installation/update steps required for this code to work properly (drush commands, scripts, configuration updates) are documented in the ticket.

## Checklist for the Peer Reviewers
- [x] The branch name of this PR matches the project standards.
- [ ] QA steps are followed and any changes to process are updated in the ticket.
- [x] Code Standards are followed, there are no bad practices in use.
- [ ] Config changes (if any) include only necessary changes.
- [ ] No errors in Drupal or Client Browser.
- [ ] Code has been tested locally (if possible).
- [ ] Following AC and Testing Steps verify changes work as expected.
- [ ] There are no known side-effects outside of expected behavior.

PR #392, opened 2022-07-27
USAGOV-29 - Update Twig Layout

thetypicalprogrammer:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-29


thetypicalprogrammer:
#386 should be merged in before this is

akf:
Does this incorporate some of https://github.com/usagov/usagov-2021/pull/386 ? It looks that way at a quick glance.

config/sync/core.base_field_override.node.state_directory_record.promote.yml
```diff
@@ -0,0 +1,22 @@
+uuid: fec2315e-00a5-475b-98a7-f112199f2abe
```
akf: I think all these yml files are (or should be) in the pull request for USAGOV-30, right?

web/themes/custom/usagov/templates/block--usagov-local-tasks.html.twig
```diff
@@ -0,0 +1,65 @@
+{# {% set site_url = url("<current>") %}
+{% if 'states' in site_url['#markup']|render|render %} #}
```
akf: While I'm going to suggest you go a different direction entirely, I wanted to point out that even this amount of digging around to find a value is a sign that we should look for another way -- either a pre-process hook, or, in the case of restricting text to a specific page, using one of Drupal's built-in mechanisms for that. 

Also, I don't think you meant to leave this commented out. :-)

web/themes/custom/usagov/html-blocks/footer-primary.html
```diff
@@ -16,7 +16,7 @@ <h4 class="usa-footer__primary-link">Government Information</h4>
                     </li>
 
                     <li class="usa-footer__secondary-link">
-                      <a href="/federal-agencies">Directory of U.S. Government Agencies and Departments</a>
+                      <a href="/agency-index">Directory of U.S. Government Agencies and Departments</a>
```
akf, line 19: Thank you!

web/themes/custom/usagov/sass/_state-directory.scss
```diff
@@ -1,18 +1,20 @@
 @use "usagov-variables" as *;
```
akf, line 1: I'm not strong in CSS, so I'm going to leave this for someone else to review.

web/themes/custom/usagov/templates/field--node--state-directory-record.html.twig
```diff
@@ -0,0 +1,88 @@
+{#
```
akf: I wonder whether we can use something like {% include directory ~ 'field--node--directory-record.html.twig' %} instead of all this content. (I have not tried that before.)

If that works, I think it's a plus, but we want to guard against some future developer deleting the file that has all the content! I would create an "includes" subdirectory under templates, move the content to a file there, and then make both of the field--node--etc files include it. That should be a pretty good clue.

web/themes/custom/usagov/templates/block--usagov-local-tasks.html.twig
```diff
@@ -0,0 +1,65 @@
...
+<p class="usa-intro">
```
akf: I would remove this from the twig file and put it in the "Intro" on the content node that winds up hosting this. That will make it easy for content editors to tweak the text in the future if they want to.

akf (changes requested):
Okay, I've thought a bit about how to approach the drop-down menu that you currently have in block--usagov-local-tasks.html.twig. That still feels like we're wedging it into a place that's meant for something else. See https://www.drupal.org/project/local_tasks_blocks about that. 

I think what you should do is create a "Custom Block," and add it to the Block Layout in the "Content" section. Once you've placed the block, you can restrict it to the /states path. 

Now you have two options for the HTML content of this block: You can paste it in to the Custom Block in Drupal (choose HTML rather than USWDS, and make sure to turn off output of the Title). OR, you can make a twig file for it, because once you've placed the block, Drupal will look for a twig file based on its name to "style" the output. You can totally style that output by just putting out a batch of HTML that ignores the block's fields (in which case, it would be nice to put a note about that *in* the block's field).

web/themes/custom/usagov/templates/field--node--state-directory-record.html.twig
thetypicalprogrammer: I updated the twig files, and it seems to work as expected.

thetypicalprogrammer:
I added a custom block for State Directory for now.

thetypicalprogrammer:
I made a Content Block, but I'm wondering the best way to populate the States. Should a `View` be created as well to automate adding a new state whenever a new State Directory Content is inserted? OR is it sufficient to use hardcoded paths (hrefs) to each state (i.e. `states/california` for California)? @akf

akf:
@thetypicalprogrammer Here's what I'm thinking: by the time we go live, we should have content in place for all the "states" (including districts, territories, and tribal governments) that exist now. New ones won't be added often, and as i understand it there's a non-trivial process for that. :-) So I think it's overkill to automate our part of that (plus, rarely-used automations have a tendency to crop up needing maintenance every time they're used, because the environment around them has changed since last use). Hard-coded paths are fine.

akf (dismissed):
I've figured out why this isn't working for me. You've added a custom block (as I suggested!), and when you export configuration, the structure of the block is exported, but not its content.   So that's what's missing for me, and why the new block shows up as Missing/Broken. 

Let's check in on how this was handled for the other custom blocks.

thetypicalprogrammer:
@akf @kellyyan
This is ready to be reviewed, but `USAGOV-30` should be merged first to avoid confusion. I merged in `USAGOV-30` into this branch to avoid config issues.

IsabelLaurenceau (commented):
Drop down works/looks as requested in ticket and takes it to correct url

bobert-2 (approved):
Looks good :)

PR #427, opened 2022-08-22
hotfix-adding-views-template: adding views template back into theme

jacobaaronyeager:
The missing views template keeps breaking my box.  Dunno if this is happening anywhere else.  PRing for review.


akf (changes requested):
Hiya! You shouldn't need this file; it was replaced with views-view-fields--federal-agencies.html.twig awhile back -- so we're only overriding the field display code for the specific view that has different needs. 
Try taking it back out and rebuilding your theme registry.

PR #447, opened 2022-08-29
Usagov 394 telephone number for mobile

IsabelLaurenceau:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-394

## Description

Files changed:

- web/themes/custom/usagov/scripts/telephoneForMobileNode.js
- web/themes/custom/usagov/templates/node--directory-record--full.html.twig
- web/themes/custom/usagov/templates/views-view-fields--federal-agencies.html.twig

Testing:
** Flush all caches
** May need to run bin/gulp build to rebuild theme if changes not evident?
** If no directory exists create one with phone numbers

On /agency-index open accordion

1. On desktop phone number should display same line and as plain text (no href to telephone or underline style) 
2. If multiple phone numbers were put only first should show 
3. Shrink window to <480 to simulate mobile
4. Refresh
5. Phone number should show as a clickable <a>. If you put a description only the phone number should be clickable description should be plain text 
6. Select more info

On /agencies/{your-agency-name-here}

1. On desktop phone number should display same line and as plain text (no href to telephone or underline style) 
2. If multiple phone numbers were put all should show on their own bulleted line
3. Shrink window to <480 to simulate mobile
4. Refresh
5. Phone number should show as a clickable <a>. If you put a description only the phone number should be clickable description should be plain text


akf (changes requested):
This seems to be working on /agency-index. We also want it to work on the Federal Directory Record and State Directory Record pages. (Further, we want to do the same thing throughout the site, so if you can associate this behavior with the "Telephone number" field type, rather than with a specific page or content type, all the better. ) 

I think this is something we talked about but didn't write down. 

Both types of directory records have phone fields for Phone Number, Toll-free Number, and TTY Number.

IsabelLaurenceau:
Hey Amy, You definitely did write it down so it's on me if it isn't working. 
I thought I had addressed it with the last push but I will go through and debug tomorrow morning!

IsabelLaurenceau:
Found the bug. Used querySelector instead of querySelectorAll and in my previous test I only had one phone number.

akf (changes requested):
I've run into a new problem here; it shows up on the detail page: if a phone number field has multiple entries, they're all winding up strung together with only spaces between them. Previously, they were being output as unordered lists if multiple, and paragraphs (I think) if single. 

The trick would be to get the phoneNumberField class applied just to an element around a single phone number ... which would mean modifying fields.field_phone_number.content, and I think there might be something tricky going on there ...

akf (approved):
Looks great!

PR #472, opened 2022-09-15
Usagov 352 synonyms

IsabelLaurenceau:
Created a new entity type agency synonym which is shown in agency view. Synonyms display and link to parent information. 

https://cm-jira.usa.gov/browse/USAGOV-352

## Description

Created a new entity type: Agency Synonym:
- - Agency Synonym has one field: Agency Reference. 
- - - Agency Reference is Required Field Type: Entity Reference -> Content limited to 1
- - - Machine Name:  field_agency_reference	
- - - Reference type -> Reference method: Default 
- - - Reference type -> content type: Federal Directory Record

Changed Federal Agencies View: 
- Added filter criteria content type = agency synonym
- Added new filter group & Put new filter in it’s own group separate groups with OR

Created new file usagov_directories.module:
- Pre render hook for agency view
- - For every result that is a synonym; shallow copy it's parent agency and display that information instead

Testing:
1. Rebuild of theme registry is required 
2. Flush all cache 
3. Create a couple new agencies
4. Create a few synonyms for said agencies (I like to also test synonyms that don't start with the same letter as parent agency)
5. See if synonym is displayed in view
6. See if synonym correctly displays parent information  

Sample Testing Inputs:

- Agency 
- - Synonym 1
- - Synonym 2 

- Agency Placeholder
- - AP
- Food and Drug Administration
- - FDA
- - Foods
- - Eats and Meds
- Census Bureau
- - Bureau of the Census
- - Census
- - U.S. Census Bureau 


akf:
I see a "body" field on Agency Synonym -- you probably don't mean to have that?

akf (changes requested):
I love it! And yet I have changes to suggest. Most of this is stuff I missed on my first go-around with the Federal Directory Records, along with some things I just forgot to mention. 

These are all for the new agency_synonym content type: 
* Please change the name from "Agency Synonym" to "Federal Directory Synonym." This will make it more obvious that they're connected in the CMS. It will also put Basic Page at the top of the list, which I think is nice.
* Optional: You could change the machine name as well, if that doesn't feel onerous. If you do, either "directory_synonym" or "federal_directory_synonym" would be appropriate. You'll note that I created a "Directory Record" and by the time I realized I should put the word "Federal" on it, it seemed like changing it would be more trouble than it's worth.
* Field changes to this content type: 
   - Remove the "Body" field. I think it's there by accident?
   - Change the default setting for "Promoted to front page" to false. This is on the "Edit" page for the content form itself, on the "Publishing Options" pane. 
   - In the same neighborhood, go into "Language Settings" and change the default to English (rather than site's default language, which is also English), **check the boxes** "Show language selector on create and edit pages" and "Enable translation" -- those checkboxes are what makes the language selector available. 
   - In "Display Settings," un-check "Display author and date information." This is just for consistency, seeing as we never actually display these directly. 

Finally, with the language selector available, you'll need to edit the Federal Agencies view to show only the English synonyms. (We are very close to wanting to create a view for the Spanish directory records.) There's **no** need to have a "language toggle" for synonyms, by the way.

IsabelLaurenceau:
I have a question for the last one. My approach would be to check which site we are on from the path (English or Spanish). Then in the pre-rendering hook, if we are on the English site check if the synonym language is set to English, and if so, show the synonym but if the synonym language is set to Spanish remove it from the results array. Is this what you meant? 
 
I am not sure what the structure will be like for the Spanish directory records. Is it a different content type or is it the same content type just with the language selected to Spanish? My guess (and what I assumed for the approach above) is the latter.  This would also mean keeping the language toggle for the synonyms.

akf (changes requested):
One more pedantic thing: you've hidden the Body field, but not removed it. You can remove it on the "Manage fields" tab. 
(Unless I'm missing something and you think we want to keep it for some purpose?)

IsabelLaurenceau:
No I definitely meant to remove it. I was thinking I may have missed it before but I just went through Agency Synonyms -> Manage fields -> Body -> Delete. I noticed on the three configs I just pushed that using the delete just writes it as hidden on the config; but it is gone in the cms. I'm not sure if that means it is still not actually deleted?

config/sync/field.field.node.agency_synonym.body.yml
```diff
@@ -0,0 +1,23 @@
+uuid: a5f311f3-f5c0-4dc2-8e22-40a1cf241db4
```
akf: I think this is the file that needs to go away in order to remove the "body" field. Here's my process: 

1. I updated with your latest code and noticed the "Body" field was still visible in the CMS. 
2. verified that my config was synchronized. 
3. In the CMS, edited the agency_synonym, deleting the "Body" field.
4. exported config (full export) 
5. deleted all the .yml files from my config/sync directory
6. moved the export file config-[...]-[date].tar.gz into config/sync, and extracted it. (Then I delete the export file.) 
7. Use "git status" to show what changed. 

All I see is that this file was deleted. 

I go through these steps, or something close, just about every time I commit config changes. There are other approaches, but this works well for me.

akf (changes requested):
I think there's just that one file to delete, and this will be good. Thanks!

config/sync/field.field.node.agency_synonym.body.yml
IsabelLaurenceau: Thank you! That showed the deleted file in my changes too. I was also exporting the full config but i was just dragging the new files into the old directory and replacing all. Because of that the unused file was still in it.

akf (approved):
Looks good!

PR #526, opened 2022-10-17
Usagov 497 username enumeration

ghost:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-497

## Local environment usage - in order to use drupal logins in your local environment:

After a build with this PR, the local environment should be ready to go - the local settings are handled by settings.local.php.    

## Local environment testing:

After a fresh build (without the db import - this is important, as our standard db will overwrite the default values!), the following should be true:

The URL <site>/user/registration should not be recognized.
The URL <site>/user/reset should not be recognized.
Test that you can toggle the login settings back to their production defaults by doing the following:
comment out the following line in settings.local.php:
$settings['usagov_login_local_form'] = 1;

Run bin/drush cr
Make sure http://localhost/user/login  has only the Secureauth button present
Re-comment that line, clear cache again, and make sure your standard login fields are present.

## Description

Modify usagov_login module such that the standard drupal login fields are not active
Modify user.settings.registration such that anonymous users cannot request a new account

- [X] A link to the JIRA ticket has been added above.
- [X] The JIRA ticket identifies the desired result of this change.
- [X] The JIRA ticket contains clear acceptance criteria.
- [X] The JIRA ticket contains clear testing steps.
- [X] The JIRA ticket contains a description of "Done".
- [X] Any preparation/installation/update steps required for this code to work properly (drush commands, scripts, configuration updates) are documented in the ticket.


akf (dismissed):
I used "pr-review" to pull this down as a fresh install. I can confirm that only the "Login via SecureAuth" button is offered with the (new) default configuration. 
For local development work, I ran into a snag: after I uncommented the two noted lines in settings.local.php and cleared my cache, the login form as the "Log in" and "Create new account" buttons, but lacks the username and password fields I would actually use in order to log in. 

(Maybe my usage pattern is unusual: when I put a fresh database in use on my local dev, I use "drush uli" to log in as root, then I find my account and put a password on it so I'll be able to log in. Thereafter, I log in with my regular account. If I were testing access I might put passwords on other accounts to test them, or create accounts for testing, but usually that's all.) 

So I would like this to offer a way for the local developer to actually log in. Maybe there's a step I missed? I'd also like to see instructions for making the local changes added to the README, probably near the instruction about using "drush uli".

akf:
I should add -- I don't see any reason to need the "Create new account" or "Reset your password" buttons, even for local dev. We would only be creating new accounts by logging in as a privileged user and adding them via the admin interface.

ghost:
> I should add -- I don't see any reason to need the "Create new account" or "Reset your password" buttons, even for local dev. We would only be creating new accounts by logging in as a privileged user and adding them via the admin interface.

You are correct Amy.  The user.settings.register value can stay 'admin_only'.     This has been explicitly set in settings.local.php, because (even though user.settings.yml has register set to 'admin_only') my local builds still have that setting showing up as 'visitors_admin_approval'.

Still looking for why this is happening...

ednark-gsa (approved):

PR #623, opened 2023-01-03
Usagov 646 move elected officials

IsabelLaurenceau:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-646

## Description

Appended to hook usagov_theme_suggestions_alter found in usagov.theme to allow us to create twig templates for code that was in the body field in the cms. Essentially this hook gets the current pages alias and appends it to the string 'field__node__body__basic_page_'. The new string is then added as a twig template suggestion. 

Item(s) of note: 
- For some reason the string in the theme file has to use underscores NOT dashes. The actual twig template can use dashes but it took hours of debugging to find that Drupal will only find the template if the hook uses underscores. 
- A previous implementation of this (see commit history or [Branch 644)](https://github.com/usagov/usagov-2021/tree/USAGOV-644-Move-States-From-CMS) used an include statement in the field--node--basic_page. Though this allowed the English and Spanish templates to be in the same file (which this implementation does not) it lead to a cluttered basic_page template. 
- As mentioned above, English and Spanish twig templates have to be separate files. The hook works by taking the current path alias. Because of this the English and Spanish pages have different hooks. 
- For the hook to be called the body in the cms cannot be empty. Any information in the body will be overwritten by the twig file but the hook is only ever called if there is data to begin with in the body field in the cms. 

### To Test
1. You may need to rebuild your cache/ theme registry since there are two new files. 
2. Once you do navigate to /elected-officials. 
3. If you do not have twig debugging set up:

- Navigate to web/themes/custom/usagov/templates/field--node--body--basic-page-elected-officials.html.twig
- Add a print statement of your choosing
- Validate you can see the print statement and that the body is being 

3. If you do have twig debugging set up:
- Inspect the body element 
- Verify the element is being populated from the template (i.e you should see:  )  


web/themes/custom/usagov/usagov.theme
```diff
@@ -50,4 +50,16 @@ function usagov_theme_suggestions_alter(array &$suggestions, array $variables, $
       $suggestions[] = 'views_view_list__sub_children';
     }
   }
+  if ($hook == 'field') {
+    $object = $variables['element']['#object'];
+    if (get_class($object) == 'Drupal\node\Entity\Node') {
+      if (in_array('field__body', $suggestions)) {
```
akf: Suggestion: $suggestions will also include "field__node__body" here. How about replacing all this with: 
`if (in_array('field__node__body', $suggestions)) {`

I could be all wet, but I think that will catch the same cases and eliminate the need to explicitly check that the current object is a Node.

akf: > For the hook to be called the body in the cms cannot be empty. Any information in the body will be overwritten by the twig file but the hook is only ever called if there is data to begin with in the body field in the cms.

This is interesting! At minimum, we should take advantage of that by replacing the content in the body with a note like "DO NOT EDIT: content supplied by twig file."

akf (commented):
I just have that one suggestion for a change (although I'm mulling over a possible enhancement).

akf:
#626 is a riff on this work. I thought it might be better to have an additional "custom" field that would be rendered after the Body. That way, content editors can still get at the intro to the form (in this case). What I've done there is not complete (for one thing, I didn't add the custom Spanish content) nor is it fully tested, but take a look and see if you like the idea. 
Armed with your work so far, the only thing I stumbled over was getting the content to *not* display if the "Custom Twig Content" checkbox wasn't checked. Drupal tried to trick me here -- the behavior is different if you render the page having never touched that field, versus having turned it on and back off!

IsabelLaurenceau:
Hello! I have implemented the changes and added the Spanish Twig file. I also moved some of the styling into the sass file. 

You'll see that I put the alert box back in the twig file (I noticed that you had taken it out and I assume you were including it in the body field). My thinking here is that the alert banner should be it's own field in the cms. From what I saw we are currently only using it on the elected officials page but if that changes we should create a field so content can change the content but the div and classes can be in a template. This also allows places the alert back in the same div as the form which it previously was. If you agree I can create another ticket to add the alert field in.

akf:
I'm not wild about adding another field to the CMS. I think that if someone were adding an alert like that today, they'd use the USWDS paragraph integration to put it in the Body field. Maybe run that idea past Chris? 
My goal was to keep as much text in the CMS where it could be edited, but I hadn't thought about the div structure. 
I do still like having an explicit indication (checkbox) that there's custom content that isn't appearing in the CMS form.

IsabelLaurenceau:
I didn't consider them using the paragraph component integration. I tested using that to create the alert and the only change I had to make to mimic the style is include the margin-top-1em utility in the surrounding div. So I think that is a good approach.

akf:
Hi! I've just noticed that this is still a DRAFT pull request, which is part of why it fell off my radar. Want to merge dev into it and make it not-a-draft? I promise to take a look.

IsabelLaurenceau:
I switched this from draft but I realized merging dev into it may take some work (since I tried merging dev into 644 and everything got mad at me) so I am moving it back to draft. I'll try to have this back to ready for review by EOD -- alright updated, fixed and ready for review!

akf (approved):
Looks good!

PR #659, opened 2023-01-31
Usagov 694 life events new name

IsabelLaurenceau:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-694

## Description

Changing "Life experiences" on the homepage to "Life events" and updating the link to point to /life-events.
Enables the redirect module.

### Testing
1. On /life-experiences change the URL alias to /life-events
2. Under URL redirects click Add URL redirect
3. In Path add /life-experiences
4. In To add /life-events
5. Save 
6. Test that localhost/life-experiences redirects to /life-events


akf (commented):
I've tried it out and it's looking good so far. Running Tome now to see whether we get something there; I'll report back tomorrow. 
We might want to use/recommend using entity references rather than paths for the redirect target. I found that if I do that (that is, type in "Life Events" and select the match), the redirect appears when I go back to edit the Life Events page, whereas if I just type "/life-events", it doesn't. 
I'm curious whether choosing a language for the redirect actually does anything?

akf (commented):
Do you have changes to config/sync/core.entity_form_display.node.basic_page.default that aren't checked in or weren't exported, by chance? I see differences between what's on the file system and what's "active" on my local dev now, and they look like they're related to the redirects. I suspect enabling the module added them, so there's a bit of a chicken-and-egg thing happening. 

I'm also seeing a stubborn difference in views.view.redirect, in that my system wants to put things in a different order than yours did. Neither is "wrong," but it would be nice to know why. Probably not worth spending time on, though. If we keep seeing it as a difference when we deploy, we'll update it from a different system!

IsabelLaurenceau:
Updated! I think it may have had to do that I changed those files for the custom fields branch so when we updated from dev there was an inconsistency? The reason why I'm thinking that is because on that branch I did change the order of somethings (such as moving hide page intro under page intro). I'm not 100% sure because I don't see how it was working without them but anywho the config files are updated and pushed.

akf:
Yep, that took care of the inconsistencies! The redirect module must have added the changes for a basic page automatically, because I saw them as updated on my site but they weren't committed. I didn't get the wizard page changes though, maybe because I didn't edit a wizard page. 

Tome created an HTML file for life-experiences with a meta redirect to life-events, so that's working too.

akf (approved):

PR #772, opened 2023-03-27
USAGOV-607: slash after /es

akf:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-607

## Description

Analytics wants us to link to "/es/", not "/es". This adds a PagerPathSubscriber that is invoked during Tome static site generation; it looks for anchor markup with href="/es" or "/es?[something]" and inserts the trailing slash. 

You can test this on local dev by running static site generation just long enough to generate the index.html page. That page already has a link to "/es" (in the language toggle); you can add '<a href="/es?foo=bar">test link</a>' to the Body of the home page to get an example of the latter form. 

To test: 
* Empty your html directory
* Clear all caches. 
* Navigate to /admin/config/tome/static/generate in the CMS and submit the form
* Look for an index.html file in the html directory. When it exists, you can inspect the result. 

Believe it or not, this seemed like the easiest way. 

The change to usa_twig_vars.module eliminates a warning message I noticed while working on this. 


IsabelLaurenceau:
I'm not sure if this is what we wanted or not. Line 9 in the generated index.html file becomes "href="http://127.0.0.1/es//" 
The double ending // is what I am not sure about.
Line 177 becomes `<a href="/es/" class="language-link" hreflang="es">Español</a>`

akf:
That is not what we want! Thank you for checking and calling this out. (It's also not what I was seeing ... frustrating.)

akf:
So, my previous work to append / to the "alternate" link for /es looked great in the CMS, but two // were being appended for the static site. Honestly I'm not sure why, because that code explicitly checks for an existing /, and the PagerPathSubscriber is only looking at anchors, not <link tags ... I think. But I took the bit of code for the "alternate" link out and I'm getting the right canonical and alternate URLs, as well as the href re-writing within the page. 

I will have to double-check this result on dev when it gets there, but please check it out!

akf:
Er, someone should check it out. I remember you're out tomorrow, @IsabelLaurenceau !

ghost (approved):
Approved

PR #778, opened 2023-03-28
USAGOV-120-internal-redirects: Add internal redirects for cutover

akf:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-120

## Description

Add a set of redirects from mothership URLs to their beta equivalents. I've added these for both the CMS and the WWW server, but we will probably want to remove them from the CMS block for reasons I'll get into below. 

There are three general categories of redirects in the internal_redirects.conf file: 
* Federal and state directory redirects: These are all permanent (status 301) and several have pattern matches. The order of these matters a bit, as nginx will take the first match. 
* Temporary redirects. We plan to remove these as content is added. They (should) all end with the "redirect" directive instead of "permanent." 
* Individual permanent redirects: These match full URLs (no patterns). 

The redirect from /es to /es/ is special and goes only in the WWW config, because Drupal is redirecting /es/ to /es. So we won't be able to test that one on local dev. 

I've put these in the CMS configuration so they can be tested on local dev, but we should remove that before merging because: 
* We will be adding content to replace the temporary redirects, and having the redirects on the CMS is going to be a hindrance when we do that. 
* These could mask erroneous links in content. 

## To test: 
You'll need to run "bin/build" and then "docker compose up" to get the rules in place. Then spot-check! (Shoshana has a "magic spreadsheet" she can use to validate all the redirects once they get to a public server.) 

You might want to refer to https://docs.google.com/spreadsheets/d/1D-vmq6-Rek99oTOBXXZwb7YOo7ppy5MQ4DIE7i_RHgg/edit#gid=2047391650

Keep in mind that the A-Z pages have query strings ("?letter=a") on the cms, and path parts ("/a") on www. The redirects only know about the path-part version. 


ghost:
Hi @akf 

The redirects I tested are working OK in a local build.  Is there anything else you'd like me to test before approval?

akf:
Thanks! I'll remove the rules from the "CMS" server.

jacobaaronyeager (approved):

PR #817, opened 2023-04-04
USAGOV-866-update-tome-for-cutover

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-866


akf (changes requested):
I forgot a piece here, which is why this turned out to be so small. 

/etc/periodic/15min/generate-static-site (in .docker/src-cms) needs an update too. If we can verify that ${WWW_HOST} will have the right thing in it, we can use that and merge this in before cutover. If that's a problem, we should change it to www.usa.gov and ... I'm not sure about merging it before cutover. I'm going to ask the analytics team if that would be bad. 

So I think the more interesting task here is looking into whether we have WWW_HOST getting populated correctly for this. Take a look at the change I made to scripts/bootstrap.sh on https://github.com/usagov/usagov-2021/pull/797/files

I'm going to close that pull request (since I took the other approach with nginx config), but that bit about setting WWW_HOST to either www or beta (not both) is probably valid here.

akf:
Update to my last comment -- I just confirmed with DK that as far as he knows, it would be fine for us to have "www.usa.gov" showing up in things like canonical URLs on beta.usa.gov, as long as beta's putting out the "noindex" header (which it is). So I think we can make this whole thing a lot easier on ourselves by starting to generate the static site pages with "www.usa.gov" at any point now. Then we don't need to worry about how long it might take to get it updated on cutover day.

ghost (dismissed):

akf (approved):

PR #855, opened 2023-04-13
USAGOV-266-fix-aws-sync: add --size-only to stop uploading unchanged files

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-266


akf (changes requested):
I don't think this is a good idea. I hope Dan is wrong about "aws s3 sync" only using file sizes and not hashes to detect changes on Linux, because I can think of a lot of ways an HTML file might be changed without changing its size! 

What might be a good idea is to change this line a bit further up: 

cp -R /var/www/html/* $RENDER_DIR

to 

cp -Rp /var/www/html/* $RENDER_DIR

... which will maintain modification time. We would still be refreshing all the timestamps when we deploy.

akf:
Following up, what documentation I've found says that aws s3 sync uses file size and timestamp (not md5 hash), and this isn't just a linux thing, it's how the aws s3 sync command works.

akf:
One more thing -- we could stand to add the "-p" flag to all of the cp (not aws s3 cp) commands in tome-sync.sh.

jacobaaronyeager:
@akf Updated

akf (approved):
This looks entirely safe to me.

PR #892, opened 2023-04-26
USAGOV-191-enable-opcache

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-191


jacobaaronyeager:
I can't find any reason to not have opcache enabled on local, so I'm just turning it on.

akf (changes requested):
Okay, it looks like we *do* need to turn this off on locals. The problem with opcache on a developer's system is that it caches php code (at a level below drupal's caches), so the developer would have to restart php each time they make a code change if they want to see the result. 

Here's what I came up with as a simple change to see whether caching would break my developer flow: 

1. Load the home page and view source. Look for `<meta name="description"` -- we populate a string for that value via the usa_twig_vars module. 

2. After line 80 of web/modules/custom/usa_twig_vars/usa_twig_vars.module, add a line overwriting the value of `$usa_twig_vars['meta_description']` like so: 

```
          $usa_twig_vars['meta_description'] = usa_twig_vars_get_meta_description($node);
          $usa_twig_vars['meta_description'] = 'WE ARE HERE';
```
3. Re-load the home page and observe the value in `<meta name="description"`. With opcache disabled, it will show "WE ARE HERE" (maybe you need to do a hard refresh, maybe not). With it enabled, it will still show the original text and no amount of cache clearing will change it.

akf (changes requested):
Not quite there yet. I learned a bit testing this, and I will share! All testing was local. 

First, I ran bin/bootstrap. I saw the message "OPCache is already disabled.", which makes sense because I had built my site with code that specified that by default. 

Next, I shut down my docker containers, ran "docker system prune --all," and ran bin/init. I then checked /etc/php8/php.ini on my cms container, and found that it had both opcache.enable and opcache.enable_cli set to 1.

Thinking that bin/init hadn't run bootstrap (I am unsure about this; local dev CMS container has CF_INSTANCE_INDEX set to 1 and bin/bootstrap says it would run automatically only when CF_INSTANCE_INDEX is 0), I ran bin/bootstrap. And I saw this message: 

```
Disabling OPCache...
/var/www/scripts/bootstrap.sh: line 145: sudo: not found
OPCache disabled.
```

That last line is a LIE, opcache.enable and opcache.enable_cli are still set to 1! 

I think we should turn off opcache.enable_cli as well as opcache.enable. 

CF_INSTANCE_INDEX is set to 1 in docker-compose.yml. Looks like it's been that way a long time; my guess would be it's intentional so we don't run tome automatically, but I'm just guessing.

jacobaaronyeager:
updated and tested.  let me know if there is something that doesn't work

akf:
Looking much better. After pruning all my images and running bin/init I have this in php.ini: 

```
opcache.enable=0
opcache.enable_cli = 1
```

Do you think we should turn off opcache.enable_cli on local dev as well?

akf (approved):
Looks good. I now have this locally: 

```
opcache.enable=0
opcache.enable_cli=0
```

PR #901, opened 2023-05-01
Usagov 244 run tome more

IsabelLaurenceau:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-244

## Description

Made the new directory 1min under .docker->src-cms->etc->periodic. 
Also changed calls in docker-compose.yml and static-site from 15min to 1min. 


akf (changes requested):
I just checked the crontab on cms-dev and it contains entries for 15min, hourly, daily, weekly, monthly, and 30min -- but nothing for 1min, so we will need the crontab updated as well. I'll put some more notes on the ticket.

.docker/Dockerfile-cms
```diff
@@ -220,8 +220,8 @@ COPY bin/src/newrelic.sh /local/newrelic.sh
 RUN . /local/newrelic.sh
 
     # cron setup
-Run mkdir -p /etc/periodic/30min \
-    && echo "*/30    *       *       *       *       run-parts /etc/periodic/30min" >> /etc/crontabs/root \
+Run mkdir -p /etc/periodic/1min \
+    && echo "*/1    *       *       *       *       run-parts /etc/periodic/1min" >> /etc/crontabs/root \
```
akf: Please change the */1 to simply *

* in the minutes column means every minute, so we don't need the /1 as well. I think I pointed you toward using /1, but we don't need it. I'm not sure whether it hurts anything. 

I became fixated on this because I wasn't seeing the log entries, nor anything in /tmp/tome-log on my docker cms container, and I was looking for the problem in the crontab file. It turns out the problem was actually that the generate-static-site script checks that $CF_INSTANCE_INDEX is 0, and on local dev it's 1 (probably on purpose to prevent running tome!) I changed that and was able to confirm that the crontab with * in the first column does kick off tome. 

I'm not seeing additional files created every minute (I expect a short file containing a "tome is already running" message), but I'm now unsure what to expect in a local setup for this job. I'll take a look later on and report back.

akf: Confirmed: tome is running every minute. We will want to pay attention to dev and see if we need to ramp up our efforts to clean the tome-log directory.

akf (changes requested):
Just one change -- removing the /1 from the crontab entry.

akf (approved):

PR #929, opened 2023-05-05
USAGOV-816: remove unintended views

akf:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-816

## Description

Disabled or restricted (to logged-in users) views that produce pages we don't want on the static site. Since one of these was responsible for creating the rss.xml page, I also removed rss.xml from the tome_static_path_exclude list in settings.php.

This is almost entirely a config update, so you need to import config to test it. 

What to test? I don't know ... key things I looked at are: 

* There is no longer a page at localhost/rss.xml
* localhost/agency-index-view is available only when logged in (we need this "page" view because without it, the "glossary" block for the agency-index pages raises an error
* localhost/agency-index pages still look good

### DIFF
--- a/config/sync/views.view.federal_agencies.yml
+++ b/config/sync/views.view.federal_agencies.yml
@@ -21,6 +21,10 @@ dependencies:
     - field.storage.node.field_zip
     - node.type.agency_synonym
     - node.type.directory_record
+    - user.role.content_administrator
+    - user.role.content_editor
+    - user.role.content_manager
+    - user.role.site_admin
   module:
     - link
     - node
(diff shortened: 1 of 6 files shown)


IsabelLaurenceau:
Should both http://localhost/agency-index-view and http://localhost/rss.xml result in this? I assumed rss.xml would result in this but not agency-index-view? 

Screen Shot 2023-05-08 at 7 13 21 AM

Screen Shot 2023-05-08 at 7 15 07 AM

akf:
Your expectations match mine -- when I visit http://localhost/agency-index-view I see content. I think I see why though: 

I see you're getting "You are not authorized to access this page" even though you're logged in. I made that page available to Content Editor, Site Admin, Content Manager, and Content Administrator ... so even if you're not logged in as a Site Admin I'd expect you to have access. 

I logged and used "drush uli" to log in as the admin user, and lo and behold, I see what you're seeing! So the views module must not automatically give access to everything to the admin user. I think that's weird, but it's okay as long as the pages at /agency-index work (and work for not-logged-in users).

IsabelLaurenceau:
Awesome! The /agency-index page does work for not-logged-in users so I'll approve and deploy it to dev.

IsabelLaurenceau (approved):

PR #947, opened 2023-05-10
USAGOV-872: tome redirect redux

akf:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-872

## Description

These are some changes to our static-site generation tweaks. This started off as another attempt to resolve the problem where a redirect page is generated for /es, but it didn't pan out that way. Nonetheless, this provides some performance improvements and ensures we don't publish /node paths. 

* Remove some of the logic for removing "/es/" from paths in the TomeEventSubscriber class; it didn't work.
* Separate "excluded directories" into a separate setting variable usagov_tome_static_path_exclude_directories. Previously, I was interpreting anything in tome_static_path_exclude that ended with / as a directory, which meant I couldn't specify "/es/" as a path to exclude there. Unfortunately, specifying /es/ turned out to be bad news. 
* Added "/node" and "/es/node" to the excluded directories. Federal directory synonyms have been populating these paths, and this is the easiest way to exclude them. 
* Convert "placeholder paths" like "_entity:node:en:1234" to canonical paths, to make the directory exclusion work. (Tome's built-in path exclusion happens later in the process after something like this has already been done, but we can't inject changes there without patching. This does not duplicate work, since we save the path changes; it just moves it forward in the process.) 

The substantial change here is the last bullet point. It cuts about 200 pages (node entities) out of processing and shaves off roughly a quarter of the time it takes to do a full static site regeneration, on my local setup. 

### Testing

Mainly, review of the code changes is what's needed. 

It probably makes sense to verify that Tome runs for you with these changes! Switch to this branch and do a "drush cr" -- this seems to be what's needed to pull in the new "event subscribers." In a terminal in your cms container, run "/var/www/scripts/tome-static.sh http://127.0.0.1" to run tome. 

Anything further is going to be time-consuming, but here's one of the things I did to test: I ran tome as above from the "dev" branch, then moved the "html" directory it contained to "html.bak". Then I switched over to this branch, cleared cache, and ran tome again. I then ran this command to identify files that were in one of those directories but not the other: 

`diff -rq html.bak html | grep Only`

We *want* to see that the new code doesn't produce some directories, like node, es/node, an taxonomy. But generally we should see the same stuff. 

Diff of whole files is not helpful, unfortunately, because of the unique query strings Drupal appends to links to assets on each run.


web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/TomeEventSubscriber.php
```diff
@@ -2,46 +2,59 @@
...
+  protected $entityTypeManager;
+
+  /**
+   * The language manager.
    *
-   * @param \Drupal\tome_static\Event\CollectPathsEvent $event
-   *   The collect paths event.
+   * @var \Drupal\Core\Language\LanguageManagerInterface
    */
-  public function excludeEsSlash(CollectPathsEvent $event) {
```
akf, line 29: This was intended to solve the "es redirect" problem but it did not. Deleting it in favor of using an excluded path.

IsabelLaurenceau:
I read the code changes and they made sense and seemed correct to me. 

So I tried running tome twice in my cms terminal. The first time I hadn't deleted the s3/.minio.sys directory. I ran it again after I deleted it and it still produced no static site files. 

Screen Shot 2023-05-24 at 10 34 26 AM
Screen Shot 2023-05-24 at 10 36 19 AM

IsabelLaurenceau:
So update. I tried to also run tome off dev. It again did not actually generate the static site but it didn't give me the error this time.  I also did run it twice. (The second time I deleted the minio directory again.) 

Screen Shot 2023-05-24 at 12 15 07 PM

akf:
I should have mentioned -- the files should be being saved to /var/www/html (which is also the html directory in your usagov-2021 repo). The minio directories shouldn't come into it, except that tome will be copying some image files from there if it can. There's a later step in the tome-sync.sh script that copies files to "s3" and I haven't changed that. 

Do you have files in /var/www/html? 

Regarding the error message highlighted in red, I'm surprised you don't see it when you run tome on the dev branch. We see it all the time in production. It's happening (I believe) because we have a page named "libraries" and there is also a "libraries" directory tome wants to copy over -- fortunately we don't actually use any of the files in that directory.

IsabelLaurenceau:
I don't have a /var directory at all. But when I look at the /html directory I have some files but it's only four so I don't think that's correct either. 
Screen Shot 2023-05-24 at 12 39 09 PM

IsabelLaurenceau:
Okay looks like it's working correctly. Exactly like you said in the directory produced from your branch we don't have node, es/node or taxonomy. I feel good about approving this if you don't have any objections. 

Screen Shot 2023-05-24 at 4 40 38 PM

akf:
@IsabelLaurenceau please do approve, thank you!

IsabelLaurenceau (approved):

PR #967, opened 2023-05-22
Usagov 24 phone numbers 

IsabelLaurenceau:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-24

## Description

When users are on a screen size <480  phone numbers that match the forms (X-XXX-XXX-XXXX; XXX-XXX-XXXX;XXX XXX XXXX) are clickable and start a call. Mimics federal directories functionality so only works on load not on resize. 

To test just shrink your screen size and refresh the page. You may have to update theme registry since a new js file was added. 


akf (changes requested):
This works well, but I noticed what is probably a little glitch in the footer where we have "1-844-USAGOV1 (1-844-872-4681)": What I see is that when the anchor markup is applied, the phone number moves to a new line and has a space before the closing parenthesis. Actually I think it's just a space that's being added that moves the number to a new line. 

Inspecting the markup, I see that this script has put an <a> inside an <a> (the link that encompassed both phone numbers in the footer), which isn't valid. 

I don't have a clever quick fix for this one. My preference would be to validate (somehow) that the text to be replaced isn't already within a link and skip it if it is.

IsabelLaurenceau:
I did get rid of the space, but for the footer I saw that and also saw that we don't use the phone number in the footer on prod! And the header phone number link is hidden on mobile.

akf (approved):
Good enough!

PR #996, opened 2023-06-06
USAGOV-1013 - add CircleCI steps

ghost:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1013

## Description

Add pre/post backup steps to CircleCI workflows.   These should disable/re-enable tome and maintenance mode.


bin/deploy/try-tome-enable
```diff
@@ -8,8 +8,22 @@ if [ x$1 == x"--dryrun" ]; then
   shift
 fi
 
+SCRIPT_DIR="$( cd -- "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
+if [ -f $SCRIPT_DIR/includes ]; then
+  . $SCRIPT_DIR/includes
+fi
+
+needsDisabled=0
```
akf, line 16: I don't see needsDisabled being used anywhere, other than here where it's set to either 0 or 1. Am I missing something?

akf (changes requested):
This is all working well. I've realized that I think we're missing a small piece of the backups! 

For restoration of the CMS, we need a backup of the files in s3://$S3_BUCKET/cms/public/ as well as the database. Now, technically we have that as part of the site snapshot (I think), but it's nice to just have that set as separate files.

Currently we use bin/cloudgov/files-pull to grab those files (which we also save to drive for developers to install locally). 

Would it be blowing scope to add that set of files to the snapshot process on this ticket?

ghost:
> This is all working well. I've realized that I think we're missing a small piece of the backups!
> 
> For restoration of the CMS, we need a backup of the files in s3://$S3_BUCKET/cms/public/ as well as the database. Now, technically we have that as part of the site snapshot (I think), but it's nice to just have that set as separate files.
> 
> Currently we use bin/cloudgov/files-pull to grab those files (which we also save to drive for developers to install locally).
> 
> Would it be blowing scope to add that set of files to the snapshot process on this ticket?

Good catch - I think I want to add another "bundle" to the process (static site, db, and now public).  I don't think these are part of what we download currently in the static snapshot.  I will double check before embarking on this effort.

bin/deploy/try-tome-enable
ghost: I will take a look - it should be executing the same logic that the try-tome-disable script uses (e.g. if ssh needs enabled, then it will also need disabled).

akf (approved):

PR #1129, opened 2023-08-14
USAGOV-1100: Delete files after Tome fails, so the next run doesn't generate an incomplete corpus

akf:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1100

## Description

When tome-static.sh fails, we already note that and don't push the files to s3. With this change, we also remove the contents of /var/www/html (I'm not worried about . files). The reason is that we have now observed that a failed run with partial content under agency-index can result in the next run's skipping agency-index without showing any error -- basically Tome doesn't pick up the need to generate additional pages. 

I had initially set out to do this when tome-static.sh reported failure with signal 9, but then realized that this is probably appropriate regardless of the non-zero return status.  

To test locally: Open a couple of terminals on your cms container. In one, run: 
  /bin/sh /var/www/scripts/tome-static.sh http://127.0.0.1

Watch for progress bars. When you start seeing them, check the contents of /var/www/html in your other terminal; there should some files. Then run: 

  ps | grep tome

And look for the parent tome process, which will look like this: 

 php /var/www/vendor/bin/drush tome:static -y --uri=http://127.0.0.1 --process-count=10

Then type "kill -9 XXXXXX" where XXXX is that process's number. In the first window, you should see the script report that tome failed with signal 9, and also the new comment "Deleting Tome files to prevent inconsistency in next run" (it may take a moment to get there). The /var/www/html directory should now be empty again. 

### DIFF
--- a/scripts/tome-run.sh
+++ b/scripts/tome-run.sh
@@ -89,7 +89,7 @@ if [ "$CONTENT_UPDATED" != "0" ] || [[ "$FORCE" =~ ^\-{0,2}f\(orce\)?$ ]] || [ "
   else
     echo "Tome static build failed with status $TOME_SUCCESS - not pushing to S3" | tee -a $TOMELOG
     echo "Deleting Tome files to prevent inconsistency in next run" | tee -a $TOMELOG
-    rm -r /var/www/html/* | tee -a $TOMELOG
+    rm -rf /var/www/html/* | tee -a $TOMELOG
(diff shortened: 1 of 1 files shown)


ghost (commented):
@akf - did you not want '-rf' instead of '-r' ?

akf:
@msvgsa "rm -r" should be enough, but sure!

ghost (approved):

PR #1146, opened 2023-08-25
Usagov 826 auto delete entities

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-826


web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.info.yml
```diff
@@ -0,0 +1,5 @@
+name: 'USA Orphaned Entities'
+type: module
+description: 'Removes orphaned synonyms and language toggles when attached Federal Directory Records are deleted.'
```
akf: I had meant the bit about language toggles to apply to all content pages, not just Federal Directory Records.

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module
```diff
@@ -0,0 +1,79 @@
...
+      $query = \Drupal::entityQuery('node')
+        ->condition('type', 'agency_synonym')
+        ->condition('field_agency_reference', $deleted_id)
+        ->accessCheck(TRUE);
+      $results = $query->execute();
+      $synonyms = Node::loadMultiple($results);
+
+      if (!empty($synonyms)) {
+        $additional = "<br /><br />";
+        $additional .= "WARNING! Deleting this node will also delete the following nodes:";
```
akf: Let's use a list instead of <br>s. It will be easier to scan. 
I suggest removing the word "WARNING!" as well.

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module
```diff
@@ -0,0 +1,79 @@
...
+        $segway = "Additionally, t";
+      }
+
+      if (!empty($entity->field_language_toggle->target_id)) {
+        $additional .= "<br /><br />";
+        $additional .= $segway . "he following nodes will have their language toggle links cleared:";
+        $lang_toggle = $entity->field_language_toggle->target_id;
+        $toggle_nodes = Node::loadMultiple((array) $lang_toggle);
+        foreach ($toggle_nodes as $toggle_node) {
+          $additional .= "<br />" . $toggle_node->language()->getName() . " node: <a href=\"/node/" . $toggle_node->id() . "\" target=\"_blank\">" . $toggle_node->title->value . " (" . $toggle_node->id() . ")</a>";
```
akf: Unordered list here, too.

akf (changes requested):
I tried this out and it deleted the synonym for me, but it didn't remove the language toggle from the remaining page. 

Additionally, I'd intended that the "language toggle" removal would apply to all content types, not just directory records. 

If this were specific to directory records, I'd suggest moving it into the usagov_directories module, rather than creating a separate module. It might still make sense to move the directory-specific logic into there.

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module
jacobaaronyeager: Good call, didn't cross my mind for some reason.

akf (commented):
I am getting error messages! When the confirmation form is displayed, I see this in the error log: 

```
cms       | [25-Aug-2023 16:39:40] WARNING: [pool www] child 2045 said into stderr: "[WARNING] [php] [2023-08-25T09:39:40] Warning: Undefined variable $additional in usa_orphaned_entities_form_alter() (line 41 of /var/www/web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module) #0 /var/www/web/core/includes/bootstrap.inc(347): _drupal_error_handler_real(2, 'Undefined varia...', '/var/www/web/mo...', 41)"
```

And I get the same message when I submit the confirmation form.

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module
```diff
@@ -0,0 +1,94 @@
...
+ * Implements hook_form_alter.
+ */
+
+function usa_orphaned_entities_form_alter(&$form, &$form_state, $form_id) {
+  if (str_contains($form_id, 'delete_form')) {
+    $markup = "<div class=\"default-message\">" .
+              $form_state->getFormObject()->getDescription() .
+              "</div>";
+    $entity = $form_state->getFormObject()->getEntity();
+    $segway = "T";
```
akf: $segue?

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module
```diff
@@ -0,0 +1,94 @@
...
+          foreach ($synonyms as $synonym) {
+            $additional .= "<li>Synonym: <a href=\"/node/" . $synonym->id() . "\" target=\"_blank\">" . $synonym->title->value . " (" . $synonym->id() . ")</a></li>" . PHP_EOL;
+          }
+          $additional .= "</ul></div>" . PHP_EOL;
+          $segway = "Additionally, t";
+        }
+      }
+    }
+
+    if (!empty($entity->field_language_toggle->target_id)) {
```
akf: I created two "basic page" nodes, set the toggle on node A to node B, and when I deleted node B, I didn't get a warning about node A's toggle, nor was it deleted. And I think this line is why -- you're checking for a toggle on the node that's being deleted. 

Usually, for a pair of pages, toggles will be created in both directions, but that won't always be the case (usually because someone forgot a step). It would also be possible (I guess) for there to be multiple nodes with toggles pointing to the same node. So can we do a search based on that? 

I'm sure this is part of the reason Drupal doesn't do this automatically...

akf (changes requested):
Looking much better, but see my note about toggles that only go in one direction.

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module
jacobaaronyeager: This is def an oversight, I didn't consider this possibility.  Thankfully, it should be rather easy to find entities with any given NID and clear them.  My only worry is memory at the moment, but we'll see.

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module
jacobaaronyeager: Sorry, it's a joke, the company who made those powered scooters is spelled that way, and I've always found it funny.

akf:
> I am getting error messages! When the confirmation form is displayed, I see this in the error log:
> 
> And I get the same message when I submit the confirmation form.

Gosh, github markdown, thanks for hiding the error message I put in there!

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.info.yml
akf: Please also update the description, e.g., 
"Removes language toggles when the nodes they refer to are deleted, and removes orphaned synonyms when the corresponding Federal Directory Records are deleted."

web/modules/custom/usa_orphaned_entities/usa_orphaned_entities.module
```diff
@@ -0,0 +1,103 @@
...
+      $results = $query->execute();
+      $synonyms = Node::loadMultiple($results);
+
+      foreach ($synonyms as $synonym) {
+        $synonym->delete();
+      }
+    }
+
+    $query = \Drupal::entityQuery('node')
+      ->condition('field_language_toggle', $deleted_id)
```
akf, line 93: I saw "Warning: Undefined variable $deleted_id in usa_orphaned_entities_entity_predelete()" when I deleted a page (basic page, not directory record), and the toggle was not deleted. It looks like the fix would be to move line 78 up above line 77.

akf (changes requested):
I still hit a bug where the toggle reference wasn't deleted when I deleted a Basic Page. See the note I attached to usa_orphaned_entities_entity_predelete.

akf (approved):
Working well now!

PR #1156, opened 2023-08-30
USAGOV-1120 tome-issue-remediation

ghost:
## Jira Task

@akf  per the discussion Thursday AM, I've committed the change to 4, but with an optional environment variable override, that we can use to adjust the count w/o requiring a deployment (granted, a restage is needed).

There is error checking to prevent invalid env var values from doing bad things - Valid values are 1 thru 10.   No non-negative integers or other (string) values are allowed.   The value is adjusted to 4 if there is nothing in TOME_PROCESS_COUNT, or if the value does not pass the validations.

To test the validation, I added the following at line 17:

```
echo "TOME_PROCESS_COUNT $TOME_PROCESS_COUNT"
exit 4
```

Then, set TOME_PROCESS_COUNT to various nonsensical values and run it locally.

https://cm-jira.usa.gov/browse/USAGOV-1120


.circleci/config.yml
```diff
@@ -540,6 +540,7 @@ workflows:
                 - dev
                 - stage
                 - prod
+                - USAGOV-1120-tome-issue-remediation
```
akf: Please remove. :-)

akf (changes requested):
It's slightly annoying that we get the warning "TOME_PROCESS_COUNT '' is not a valid, non-negative integer.  Adjusting to 4" when we haven't set TOME_PROCESS_COUNT at all. How about adding the default up top: 
TOME_PROCESS_COUNT=${TOME_PROCESS_COUNT:-4} 

Also, the circleci config needs to be changed back. (I was happy to test locally and just hit ctrl-c after it got past the point of kicking off tome.)

.circleci/config.yml
ghost: These are removed, and we now avoid giving a warning if the env var is empty.

akf (approved):

PR #1163, opened 2023-09-05
Usagov 1116 stable phone numbers

IsabelLaurenceau:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1116

## Description

Previously we had two js files for telephone numbers. One for telephone numbers in the agency index (http://localhost/agency-index) and one for a phone number in the body (http://localhost/school-meals). This allows us to now only have one JS file because we changed the agency view telephone number fields to plain text fields.

## Testing Instructions
**requires config changes import** 
**requires theme rebuild** Removes and renames files 
On mobile in chrome, safari and firefox test that 
1. All phone numbers are clickable links
2. In inspector test that there aren't nested a tags (safari automatically makes numbers links so need to make sure we aren't creating a link in a link) 
3. Test that phone numbers appear in desktop on all browsers and are **not** clickable 


YaritzaGarcia (approved):

web/themes/custom/usagov/scripts/phoneForMobile.js
```diff
@@ -1,4 +1,4 @@
-var pattern = /[0-9]?[-. ]\(?([0-9]{3})\)?[-. ]([0-9]{3})[-. ]([0-9]{4})/g;
+var pattern = /[0-9]?[-. ]?\(?([0-9]{3})\)?[-. ]([0-9]{3})[-. ]([0-9]{4})/g;
```
akf, line 1: I'm looking at this because I wanted to see what you wound up doing and this pattern caught my eye: 
`[-. ]` shows up twice here and several times on the next line. I think that matches "-", any character, or " ". Probably what was meant was `[-\. ]`

(I say "what was meant" because I'm pretty sure this pattern came from old code elsewhere.) 

If I'm right about the intent, I'd be in favor of opening a new ticket to tighten this one up. It's hard for me to  imagine us having a pattern that is _not_ a phone number that would match the looser pattern, but it's liable to lead someone astray in the future.

IsabelLaurenceau: I don't think so. I wrote this regex pattern myself and I wanted that to match either the - . or an empty space so that it would match numbers: 1-123-456-7890; 1.123.456.7890; or 1 123 456 7890. The closed brackets [] say to match any character in them so that's why it's written "-. "  When I tested it with a regex tester the only characters that do match with it are those three.

akf: Ah, you are correct! Thanks for replying on that. It turns out a dot is just a dot when used within square brackets. TIL! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Cheatsheet#:~:text=Inside%20a%20character%20class%2C%20the%20dot%20loses%20its%20special%20meaning%20and%20matches%20a%20literal%20dot.

PR #1229, opened 2023-10-17
Drupal 10 config re-export 

akf:
I have a bunch of what I think are non-meaningful differences between the configuration files and the configuration in my database -- that is, I think these make defaults explicit and the like, because they keep showing up as changes after successfully applying. 

I would like to check that it's not just me before we merge this, though.


arpage:
From what I see, the following has happened:

- media loading defaults added
- 2 new modules in core extensions
- Quoting changes (single -> double)
- a change in the "third_party_settings" config format
- tooltip and time_diff defaults added

The site looks ok on my local.

arpage (approved):

PR #1424, opened 2024-02-15
Usagov 1313 sayt federal directories

IsabelLaurenceau:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1313

## Description

Create a drop down menu for federal agencies.

Screenshot 2024-02-15 at 3 33 45 PM

## Type of Changes

- [x] New Feature
- [ ] Bugfix
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

### Requires New Config
- [x] Yes
- [ ] No

### Requires New Content
- [ ] Yes
- [x] No

### Validation Steps
- You'll need to import config changes so "bin/drush cim"
- On http://localhost/agency-index you should see the search box that's been there
- Now however if you type you should start to see a list of agency suggestions appear. The search space for this list are the agency names, synonyms and acronyms we have for them in Drupal. 
- Clicking on the suggestion should take you to the requested agency page. 
- Clicking outside the suggestion box should hide the box. 

DK provided a list of the most search terms in the box. Please test out at least some of these during the review. 
Screenshot 2024-02-15 at 3 32 21 PM
Screenshot 2024-02-15 at 3 32 47 PM

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders
- Reviewed code changes
- Reviewed functionality


web/themes/custom/usagov/components/searchbar/searchbar.js
```diff
@@ -0,0 +1,93 @@
+const search_input = document.getElementById("search-field-en-small");
```
akf, line 13: Please add some comments about what this script is for and what it does for future maintainers (yes, it seems obvious right now!) 

> This bit from the original request is the type of thing I'm thinking of: 
> When someone starts typing in a keyword, the agency names are suggested in an autofill.
> 
> For example: If someone types in "internal" The auto fill should come up with "Internal Revenue Service" and take the user directly to the directory page.

Plus, I would note that it relies on the federal_directory_export view.

akf (changes requested):
This is great! I discovered, though, that it doesn't work when I'm not logged in, which means it's not going to work for the static site. 

The fix for that will be to change "Access: Role" on the view to "Access: Unrestricted". That can either be done individually on the JSON displays, or for the entire view (all displays). I think it would be fine to make the CSV reports available to unauthenticated users as well! 

We'll need to check that the static site generator actually picks up these views and includes them in its output. 

The other change I'd request (optional!) is to rename the JSON exports. What I'd like to see is: 
 * The displays are named something that indicates that their content is just for the search suggestions, maybe "Search suggestion En" and "Search suggestion Es" (or "SAYT En" etc.) 
  * The path settings similarly suggest the purpose of the data. 
 This is coming to mind because these displays rightly contain a small subset of the data of the CSV display. 

(I'm also not sure whether sites/default/files is the best spot for these. They could go under /agency-index and /es/indice-agencias, or under a named directory like data-exports.)

akf:
I triggered scan-code to make sure there weren't any new issues. The two things it flagged and "failed" on are pre-existing issues -- nothing to worry about for this PR!

IsabelLaurenceau:
> (I'm also not sure whether sites/default/files is the best spot for these. They could go under /agency-index and /es/indice-agencias, or under a named directory like data-exports.)

I decided on sites/default/files since I knew it was a readable path but also I noticed that is where benefit finder is writing their files to too. I think it makes sense that either we write all the files to this path or we make it so whenever we write files for a purpose it goes within a path linked to functionality. I just want to make sure we are consistent so we know where all the files should be found on the site.

IsabelLaurenceau:
Changes addressed!

config/sync/core.entity_form_display.node.directory_record.default.yml
akf, line 1: I scratched my head over these, but it looks like these (especially simple_sitemap) are defaults that are now explicit.

config/sync/views.view.federal_directory_export.yml
akf, line 1: Oops -- the CSV export of this view has changed. It's now Spanish only. (It's in the Filter criteria in the UI.) 

CSV export should be: 
  * both (all) languages
  * directory records only (no synonyms). I didn't think to include them the first time around, but they're not very useful as they are now in the CSV. 

SAYT exports are good as they are now, but I wonder whether "Original language" is the right field? Maybe you verified this -- I'm wondering whether, if someone creates a record with the wrong language selected and then updates it, it winds up in the correct export.

akf (changes requested):
It occurred to me to check whether there were any unintentional changes to the CSV export and ... there are. That's what needs to change, and my comments are attached to the views.view.federal_directory_export.yml file. 

I've run the tome export "long enough" on my local site to see that the JSON files are generated at the expected locations. Be aware that unless we do something special, those are probably not going to get "Content-type: application/json" headers on the static site like they do from Drupal. I don't think that's going to be a problem here; your code is just expecting a response. 

Checking the headers made me very aware that this javascript requests the same file every time I type a letter. I would think that browser caching would mitigate that, but it would be very nice if you could save the response in a javascript variable and not re-request it every time. I don't see a need to do anything fancy to make it persist across different page loads, though. 

So: 

1) Change the CSV export settings back (required)
2) Request the JSON data only once, when the first letter is typed, if possible.

config/sync/views.view.federal_directory_export.yml
IsabelLaurenceau: Changes made to the csv export! For the original language I'm going based the fact that original language refers to node.langcode which is what we use for language checks in the code.

IsabelLaurenceau: Changes made to request the JS only once!

akf: Looking good now!

akf (approved):
Works great!

PR #1460, opened 2024-03-04
Usagov 1373 rework ssg post processing

YaritzaGarcia:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1373

## Description

Modified the modifyHtml function in usagov_ssg_postprocessing/src/EventSubscriber/TomeEventSubscriber.php to use html5-php instead of DOMDocument(). This fixes the bug in dataLayer output for Spanish pages.

## Type of Changes

- [ ] New Feature
- [x] Bugfix
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

### Requires New Config
- [ ] Yes
- [x] No

### Requires New Content
- [ ] Yes
- [x] No

### Validation Steps
- Empty your html directory
- Clear all caches
- Go to http://localhost/admin/config/tome/static/generate and submit the form
- After a couple of seconds of running this process, you should be able to locate this file: html/index.html. Open the file, search for 'href="/es' and verify that the url ends with a slash 
image

- When this process is about to finish, you should have a folder titled "es" inside the html folder. Select any of those pages and inspect the index.html file (example: html/es/ayuda-estadounidenses-extranjero/index.html).
image

- Search for the word "Taxonomy_Text_1" and verify that it says "Página principal", not "P&aacute;gina principal"
image

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders
- Reviewed code changes
- Reviewed functionality


web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/TomeEventSubscriber.php
```diff
@@ -132,17 +133,22 @@ public static function getExcludedDirectories() {
...
     $xpath = new \DOMXPath($document);
     $changes = FALSE;
+    $nodes = $xpath->query('//a[starts-with(@href,"/es")]');
+
     /** @var \DOMElement $node */
-    foreach ($xpath->query('//a[(starts-with(@href,"/es"))]') as $node) {
+    foreach ($nodes as $node) {
       $original_href = $node->getAttribute('href');
-      $new_href = NULL;
+      $new_href = '/es/';
```
akf: This is a mistake -- and maybe an indication that I should have put a comment here! By setting $new_href to '/es/' here, you're changing *all* of the hrefs that start with "/es" to "/es/". For example, "/es/beneficios-gobierno" changes to "/es/". 

It would be reasonable to set $new_href to $original_href here. The reason I used NULL (and should have added a comment, probably) is that I wanted to avoid re-writing the HTML if none of the hrefs needed to be updated. Doing it this way let me check for that. (Why avoid re-writing the HTML? Mainly just being cautious in case this process might introduce unintended changes. It should help a little with performance too.)

web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/TomeEventSubscriber.php
```diff
@@ -132,17 +133,22 @@ public static function getExcludedDirectories() {
    */
   public function modifyHtml(ModifyHtmlEvent $event) {
     $html = $event->getHtml();
+    $html5 = new HTML5();
 
     // LIBXML_SCHEMA_CREATE fixes a problem wherein DOMDocument would remove closing HTML
     // tags within quoted text in a script element. See https://bugs.php.net/bug.php?id=74628
     $document = new \DOMDocument();
     @$document->loadHTML($html, LIBXML_SCHEMA_CREATE);
```
akf, line 141: Did you try using `$html5->loadHTML($html)`? 
It looks like you can replace these two lines with `$document = $html5->loadHTML($html);` 
(I'm looking at https://github.com/Masterminds/html5-php/wiki/Basic-Usage)

It seemed to me that using the HTML5 parser should be better than using php's built-in libxml, since it actually supports HTML5.

akf (changes requested):
I found a bug! (This is rewriting more of the hrefs than just "/es".) 

Other than that, this is definitely solving the diacritical problem, but I would like to know whether using the HTML5 library for parsing the HTML, as well as for writing it out, works. If it does, it's probably better to use it, even if the results are identical.

web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/TomeEventSubscriber.php
YaritzaGarcia: Oh no, I'm sorry. I was testing something and I totally forgot to fix this.

web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/TomeEventSubscriber.php
YaritzaGarcia: I did try adding that line of code to replace the two lines we have now (the commit: https://github.com/usagov/usagov-2021/pull/1460/commits/04b18402c6decca41e61f4dda202e5a9e364f315) but for some reason it didn't work even though I tried several times. I have a suspicion that this may be related to the problems I'm having with my local environment so I did a completely fresh build (again lol) to test it one more time.

YaritzaGarcia: No, just like yesterday, it doesn't work.

image

YaritzaGarcia:
> I found a bug! (This is rewriting more of the hrefs than just "/es".)
> 
> Other than that, this is definitely solving the diacritical problem, but I would like to know whether using the HTML5 library for parsing the HTML, as well as for writing it out, works. If it does, it's probably better to use it, even if the results are identical.

I just fixed this, can you review it? Thank you so much!

web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/TomeEventSubscriber.php
akf: Okay, thanks for checking! I'm surprised that just making this change on the output end solves our problem, but it definitely does.

akf (approved):
Working well now.

PR #1572, opened 2024-05-10
USAGOV-1346 call center wait times

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1346

## Description
Added infrastructure for getting estimated wait time for call center to be displayed on /phone and /es/llamenos.

## Type of Changes

- [X] New Feature
- [ ] Bugfix
- [X] Frontend (Twig, Sass, JS)
- [ ] Infrastructure
  - [X] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

## Testing Instructions
PRETESTING INSTRUCTIONS: Must add the following to /phone and /es/llamenos:
```
<div id="callCenterTime">
    &nbsp;
</div>
```

Then navigating to those pages should show the estimated wait time at the bottom of content.  If not, talk to me to work out why not.

### Requires New Config
- [ ] Yes
- [ ] No

### Requires New Content
- [X] Yes
- [ ] No

### Validation Steps
- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders
- Reviewed code changes
- Reviewed functionality


web/themes/custom/usagov/scripts/callCenterWaitTimeLookup.js
```diff
@@ -0,0 +1,74 @@
+jQuery(document).ready(async function () {
+  "use strict";
+
+  var waittime = (function() {
+    if (jQuery("html[lang|='en']").length ||
+        jQuery("html[lang|='es']").length) {
+      jQuery.ajax({
+        "url": "/wait-time",
```
akf: This would need to target the cms server specifically, once the javascript is published to the static site.

akf (commented):
We should probably go through this together, as it didn't "just work" on my system -- and I didn't hear how it went on dev. Here are some things I'm seeing: 

- I get the message "can't execute '/etc/periodic/1min/call-center-update': No such file or directory" in my logs. The call-center-update file is present, so I think the issue is that /bin/bash is not. We need /bin/sh or /bin/ash (which point to the same thing).
- You've convinced me it's okay to call an API on the CMS from the static site, but now you need to make this javascript file call the CMS host, not the static site. That might be tricky. 

I'm coming back to thinking it might be worth getting an S3 bucket set up with the right CORS headers so we don't have to use the CMS as an API.

.docker/src-cms/etc/periodic/1min/call-center-update
```diff
@@ -0,0 +1,110 @@
...
+fi
+# we might be running from a local dev machine
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+if [ -f $SCRIPT_DIR/env.local ]; then
+  . $SCRIPT_DIR/env.local
+fi
+if [ -f ./env.local ]; then
+  . ./env.local
+fi
+
```
akf: So... everything up to this point looks as if it's meant to run from a developer's terminal. In the CMS container, the S3 credentials will need to come from $VCAP_SERVICES (which is already happening in bootstrap.sh). 

I think this is probably some outdated code. I'm also thinking it might not be that hard to take the following bits and set them up in a service separate from the CMS. (Yes, I'm still on that soapbox.)

.docker/src-cms/etc/periodic/1min/call-center-update
```diff
@@ -0,0 +1,110 @@
...
+# yes I know there is a ton of repeated code here, I couldn't get functions to work properly
+if [ -e $tokenFile ]; then
+  echo "Token file exists."
```
akf: Once we have the rest of this nailed down, let's take another pass at these messages. I think we'll want to include a common string we can filter on, and clearly note ERROR versus INFO. Also, we might not need most of these at all.

web/themes/custom/usagov/scripts/callCenterWaitTimeLookup.js
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            if (minutes === 0 && seconds === 0) {
+              content = content + noneText;
+            }
+            jQuery('#callCenterTime').html(content);
+          }
+        },
+        "error": function (xhr, status, error) {
+          console.log('fail');
```
akf, line 71: Let's revisit the error messages in the next round of review.

scripts/API/call-center-update.sh
akf, line 1: I think I have confirmed, by searching for "call-center-update.sh" in the codebase, that this file isn't used and can go away. And that it's identical to the cron script.

.docker/src-cms/etc/periodic/1min/call-center-update
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
```
akf: I am still getting this message: 
```
cms       | run-parts: can't execute '/etc/periodic/1min/call-center-update': No such file or directory
```
The problem is /bin/bash -- we don't have that on the cms. I changed it to /bin/ash and got this instead: 

```
cms       | S3 service storage does not exist in space 
cms       | run-parts: /etc/periodic/1min/call-center-update: exit status 1
```

A lot of this script looks like it's supposed to run in a developer's terminal. The good news is that we already have code in /scripts/bootstrap to get the S3 environment variables (from $VCAP_SERVICES). Bootstrap would need to insert those into the call-center-update script or otherwise make them available.

web/themes/custom/usagov/scripts/callCenterWaitTimeLookup.js
```diff
@@ -0,0 +1,84 @@
...
+        'www': 'cms',
+        'cms': 'cms',
+        'beta': 'cms',
+        'beta-stage': 'cms-stage',
+        'cms-stage': 'cms-stage',
+        'beta-dev': 'cms-dev',
+        'cms-dev': 'cms-dev',
+      };
+      jQuery.ajax({
+        "url": "https://" + domainMap[domain] + ".usa.gov/wait-time",
```
akf, line 19: We don't have a case here for localhost, which is kind of sad. My dev site is trying to call https://undefined.usa.gov/wait-time 

(http://localhost/wait-time is returning the string "false", which I guess is fair since the script isn't working.)

akf (changes requested):
See my comments here -- chief among them, the call-center-update cron job doesn't work for me locally and I don't see how it could work in a cms container in our hosting environment either. 

I still think it will be easier (and better) to move this cron job to a separate server and have just one s3 bucket that serves all environments. Maybe I could work on setting up the bucket and server and you could migrate the code over. 

The code for actually making the API calls to the call center environment, and to retrieve the wait time to display on a web page, look good (although I haven't tested them).

PR #1665, opened 2024-06-26
USAGOV-1664 allow a command line argument to specify a service account name other than "cci".

arpage:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1664

## Description

1. allow a command line argument to specify a service account name other than "cci"
2. assert that it is being run in a given space, per command line argument
3. remove the cf-cli version switching code - cf v8 seems to be working now

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [x] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

### Requires New Config
- [ ] Yes
- [ ] No

### Requires New Content
- [ ] Yes
- [ ] No

### Validation Steps
- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders
- Reviewed code changes
- Reviewed functionality


bin/cloudgov/create-service-account
```diff
@@ -1,18 +1,31 @@
-#!/bin/bash
+#!/bin/env bash
```
akf: Should be /usr/bin/env (I don't have /bin/env, and we already use /usr/bin/env in many places)

bin/cloudgov/create-service-account
```diff
@@ -1,18 +1,31 @@
...
```
akf, line 2: Please, let's add some documentation while we're here. Either comments here, or a usage statement as we have in some other scripts. Something like: 

Creates a service account in the specified space (which must be current) and assigns that account roles in all spaces. 
Usage: bin/cloudgov/create-service-account space [keyname] [--reset-key]

Keyname defaults to "cci". Service account will be named keyname-service-account. 
If --reset-key is specified, the existing service account will be deleted and replaced.

bin/cloudgov/create-service-account
```diff
@@ -23,42 +36,44 @@ do
...
-cf set-space-role $SERVICE_USER $startorg tools SpaceDeveloper
-cf set-space-role $SERVICE_USER $startorg shared-egress SpaceDeveloper
-cf set-space-role $SERVICE_USER $startorg shared-egress-dr SpaceDeveloper
-cf set-space-role $SERVICE_USER $startorg dev-dr SpaceDeveloper
+
+# you must already be logged in as an OrgManager to have enough permissions
+# to assign these cross-space developer roles to the service account's user
+
+cf set-space-role $SERVICE_USER $startorg dr SpaceDeveloper
+#cf set-space-role $SERVICE_USER $startorg dev SpaceDeveloper
```
akf: We'll want to un-comment these lines before merging.

akf (changes requested):
I noted a couple of changes that we need to make. 

At the risk of opening a can of worms, now that I see that we have this script to create a service account in a specific space, and then give it permissions in all (-ish) of the spaces, it's really bothering me. "All the spaces" makes perfect sense for the CircleCI user. 

What would you think of choosing to create all the credentials in the same space (either "tools" or "prod," I prefer "tools" for this)? In that case, we wouldn't specify a space on the command line.

I poked around a little at the cci-service-account services. We have one in dev, stage, and prod -- maybe others -- but it's the prod one whose user we've set up in CircleCI.

bin/cloudgov/create-service-account
arpage: On it - thank you for looking!

bin/cloudgov/create-service-account
arpage: Should this just be setting the role for the service user within the current space, or all spaces?

It seems that since we run this script per space, we'd just want e.g. the dev user to have the role in the dev space?

arpage: @akf - my latest commit will hopefully address the cross-space access needed for CCI.

akf: @arpage I like that ... for now.

akf (approved):
Thanks for the update. Tested in my sandbox and also in prod (and then deleted the service account).

I did get an error message that "space shared-egress-dr does not exist," but I expect that space will be back!

PR #1668, opened 2024-06-26
Usagov 1561 published pages corrections redux

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1561

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

### Requires New Config
- [ ] Yes
- [ ] No

### Requires New Content
- [ ] Yes
- [ ] No

### Validation Steps
- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders
- Reviewed code changes
- Reviewed functionality


akf (commented):
I've run this once locally -- actually twice, because docker crashed part way through the first run, so the effect was as if a tome run got interrupted and then it started over. (Maybe this affects what I'm seeing.) 

I looked at the Home Page entries first. Both Home Pages appear now, but for some reason the English page appears twice. 

Also, the Toggle URL for the English home page is shown as https://www.usa.gov/ , while the Spanish home page shows "None". I can see that your new "data-type" attribute doesn't appear in the `link` elements on my home pages, so that would explain why the Spanish home page is listed as not having a Toggle URL. 

The "data-type"="altlang" solution seems like it ought to work, but a lot of pages show "None" for the Toggle URL and when I looked at one, I see the expected attribute in the page source: http://localhost/es/dias-festivos-estados-unidos

I have a bunch of duplicates -- not every page, but maybe most of them? "Why" is not apparent to me. 

I'm going to give this another try and see if anything's different, and if I can contribute any more information on why the alternate URLs aren't getting picked up properly on the home pages.

akf (commented):
Part of the issue with the home page is that the "alternate" entity is the same node. So where "data-type"="altlang" is being set, we need to be testing differently.

Scratch that -- I stepped through this with my debugger and determined that: 
1. On the home page, your code that sets `data-type` to `altlang` does run
2. I also see `data-type` is being set to `conlang` for the appropriate link

But I don't see any of this in the source of the home page(s). Conversely, on other pages, I see "altlang" but not "conlang." I've just realized there's a clause starting on line 107 of sa_translation_page_attachments_alter that might be the culprit -- it's doing a lot of the same work and I think it might be extraneous now.

akf (approved):
IT LOOKS GOOD. The home pages come out correctly, the "alternate" URLs appear to be correct (spot-checked). 

I see a few patterns that are going to look like issues -- this report is showing some things that Tome is generating on my local, which we want it to skip. I'll note those on the Jira ticket, but I think they are separate issues and not a problem with this report.

PR #1689, opened 2024-07-02
USAGOV-1629-Benefits-Category-Language-Filter

cwacht:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1629

## Description

The `Benefit Search Category Reference` View has been updated so each benefit category will display its language.
The `benefitsCategoryLanguageFilter` script converts that displayed language to a data attribute, and then filters the items by the currently selected language.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

### Requires New Config
- [x] Yes
- [ ] No

### Requires New Content
- [ ] Yes
- [x] No

### Validation Steps
- Start creating a new `Basic Page`
- Locate the `Benefit Category Search` field
- [ ] Confirm that only the english benefits categories are displayed
- Change the Language field to Español
- [ ] Confirm that only the Spanish benefits categories are displayed

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders
- Reviewed code changes
- Reviewed functionality


web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
```diff
@@ -0,0 +1,64 @@
...
+by the currently selected language. It also updates the text within the 
+<legend> element to indicate that the Benefits Categories are being filtered
+as well as which language is currently selected.
+
+It is called when the edit page loads as well as anytime the language field
+is changed.
+*/
+function filterCategoriesByLanguage(jQuery){
+    let selectedLanguage = jQuery("#edit-langcode-0-value :selected").text();
+    jQuery('#edit-field-benefits-category').children().each(function () {
```
akf: Let's add a clause here to show all categories in the case where selectedLanguage is null or empty. I know that shouldn't happen, but it seems like something to defend against, just in case.

I also have an idea that we should add a "show all" button for troubleshooting purposes.

akf (commented):
This looks nice and it worked well in my light testing. I'm just commenting here, as I'd like @omerida to review this. 

In thinking about what could go wrong, it occurred to me that it would be possible for a person to save the page in the wrong language, and then come back and fix that. I confirmed that if I start to edit the page with "English" selected, select a category, and then switch the page's language to "Spanish,"  the English category I selected is retained even though I don't see it. That's probably an unlikely scenario, given the content team's workflow, but it seems likely enough to account for. I think we should: 

* Definitely add a checkbox that lets the user display all the categories, so they can confirm what's going on. 
* Maybe extend the logic in filterCategoriesByLanguage to show all the categories if any "wrong language" category is selected.
* Ask for feedback from the content team on this specific issue. Maybe we should automatically clear any categories that don't match the page's language?

web/modules/custom/usa_admin_styles/usa_admin_styles.module
```diff
@@ -15,4 +15,5 @@ function usa_admin_styles_form_node_form_alter(&$form, FormStateInterface $form_
   // in an iframe), but is a problem in ckeditor 5 -- all preview content is within the node
   // add/edit form, so button elements in the preview will submit the form by default.
   $form['#attached']['library'][] = 'usa_admin_styles/accordionButtonSubmitFix';
+  $form['#attached']['library'][] = 'usa_admin_styles/benefitsCategoryLanguageFilter';
```
omerida: Instead of loading this library for all node forms, let's only load it for basic page node, wrapping it with an `if` like this one should do it:

```php
  if ($form_id === 'node_basic_page_edit_form' || $form_id === 'node_basic_page_form') {
    $form['#attached']['library'][] = 'usa_admin_styles/benefitsCategoryLanguageFilter';
  }
```

web/modules/custom/usa_admin_styles/usa_admin_styles.libraries.yml
```diff
@@ -6,3 +6,7 @@ mylib:
 accordionButtonSubmitFix: 
   js:
     noSubmitOnAccordion.js: {}
+
+benefitsCategoryLanguageFilter: 
+  js:
+    benefitsCategoryLanguageFilter.js: {}
```
omerida, line 12: since this depends on jquery, you should list it as a dependency. Unlikely that jquery  wouldn't be loaded on an admin page but it's helpful documentation.

```
  dependencies:
    - core/jquery
```

web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
```diff
@@ -0,0 +1,64 @@
...
+content type shows all the items by default.
+
+This script filters those categories by the content's selected language.
+
+This script assumes that each category includes its language at the end of the <label> element
+*/
+
+(function (jQuery, Drupal, drupalSettings) {
+
+    storeLanguageInDataAttribute(jQuery);
```
omerida: You can add the data attribute via rewriting the field in the view.

1. Hide the original language field and move it to first

image

2. Update the title field, under style settings you can uncheck the "add default classes" option. And under "Rewrite classes" you can put the markup you want. The replacement patterns section shows you what's available. Each field can see itself and any fields that came before it.

image

3. The edit form's checkboxes should then have the data attribute:

image

4. Update `filterCategoriesByLanguage()` to find first span in each option with a `data-language` attribute and use it:

```js
if(jQuery(this).find('span[data-language]').first().attr( "data-language" ) != selectedLanguage){
```

omerida: Then you don't need `storeLanguageInDataAttribute()`

web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
```diff
@@ -0,0 +1,64 @@
...
+
+    filterCategoriesByLanguage(jQuery);
+
+    jQuery('#edit-langcode-0-value').change(function(){
```
omerida: the `change` shorthand is deprecated along with a bunch of (maybe all?) other events - <https://api.jquery.com/change-shorthand/>

For forward compatibility, they recommend doing `.on('change', function(){`

web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
```diff
@@ -0,0 +1,64 @@
...
+as well as which language is currently selected.
+
+It is called when the edit page loads as well as anytime the language field
+is changed.
+*/
+function filterCategoriesByLanguage(jQuery){
+    let selectedLanguage = jQuery("#edit-langcode-0-value :selected").text();
+    jQuery('#edit-field-benefits-category').children().each(function () {
+        if(jQuery(this).attr( "data-language" ) != selectedLanguage){
+            jQuery(this).hide();
```
omerida: To prevent hiding selected categories, can you update this to only hide the checkbox if it's unchecked?

omerida (changes requested):

web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
cwacht: With my changes today, it should show all the categories if selectedLanguage is anything other than "true"
I've also added the `Show all categories` button while items are filtered which becomes a `Filter categories by selected language` button while showing all categories.

web/modules/custom/usa_admin_styles/usa_admin_styles.module
cwacht: Good idea. I've added this.

web/modules/custom/usa_admin_styles/usa_admin_styles.libraries.yml
cwacht: OK, do you think the other library should have this as well?

web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
cwacht: Thank you! I knew this should be possible. I think I hadn't tried changing the order of the fields... I added this change and it looks much cleaner this way.

web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
cwacht: Good catch

web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
cwacht: I did this, but it felt odd with the way the filter description was worded, so I also added extra detail to the filter description when other checked categories are being shown, and I think it helps add clarity.

web/modules/custom/usa_admin_styles/usa_admin_styles.libraries.yml
omerida: If the other library depends on jquery, it should have it

omerida:
Tested the updated functionality locally and it looks good to me.

omerida (approved):

web/modules/custom/usa_admin_styles/benefitsCategoryLanguageFilter.js
akf: I like it!

PR #1729, opened 2024-07-16
USAGOV-1674 GTM category tag

omerida:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1674

## Description

Adds benefit categories to the datalayer for Google Tag Manager for analysis
## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [x] Frontend (Twig, Sass, JS) Techincally, it's a twig update but there should be no visual changes. The source for the page changes:

image

- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

## Testing Instructions

Setup the benefits search with some tagged pages. You should see one or two new values in the datalayer for standard pages.
### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- View source for a standard page that's been tagged to one or more benefit categories. The datalayer (script tag with id=taxonomy-data)  should have entries that look like the following. Categories should be separated by a `;`
```
      "hasBenefitCategory": true,
      "benefitCategories": "Housing; Utilities",
```
- View source for a standard page that has not been tagged. The datalayer should only have one entry

```
      "hasBenefitCategory": false,
```
- All other pages should not include either of these json keys.

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


akf (changes requested):
Looks good. I see one possible gap: Currently, you're including `"hasBenefitCategory": false,` only on content types that could have a benefit category. I'll defer to the Analytics team on whether they'd prefer to have `hasBenefitCategory` in the dataLayer on all pages, or if they prefer this approach. (The conversation in Jira implies "all pages" to me, but it's not definitive.) 

Second, there is a lint error about using "true" and "false" in the php code instead of "TRUE" and "FALSE."

omerida:
@akf got clarification via Jira about how to implement and fixed the casing of bools.

web/themes/custom/usagov/templates/html.html.twig
```diff
@@ -61,6 +61,11 @@
       {{'homepageTest'|json_encode(constant('JSON_PRETTY_PRINT'))|raw }}: {{front|json_encode(constant('JSON_PRETTY_PRINT'))|raw }},
       {{'basicPagesubType'|json_encode(constant('JSON_PRETTY_PRINT'))|raw }}: {{pagetype|json_encode(constant('JSON_PRETTY_PRINT'))|raw }},
       {{'Page_Type'|json_encode(constant('JSON_PRETTY_PRINT'))|raw }}: {{datalayer_pagetype|json_encode(constant('JSON_PRETTY_PRINT'))|raw }},
+      {{'hasBenefitCategory'|json_encode(constant('JSON_PRETTY_PRINT'))|raw }}: {{datalayer_hasBenefitCategories|json_encode(constant('JSON_PRETTY_PRINT'))|raw }}
```
akf: Need a comma at the end of this line.

akf (changes requested):
I probably could have noticed this earlier, but the "hasBenefitCategory" output needs a comma at the end. 

Oddly less important, I see that some pages still don't have the `"hasBenefitCategory": false` line. It's because there's an `{% if nodetype == "basic_page" %}` clause around the text you edited in the twig file, and an `else` clause for other page types. (Example non-basic_page page: http://localhost/benefit-finder/death ). 

(Looking at *why* there's an if/else clause, I see that there's a different variable output for `Page_Type`. It took me some squinting so I thought I'd save you that at least.)

web/themes/custom/usagov/templates/html.html.twig
omerida: good catch.  

Could we refactor how this is done in the future? Doing it in the template seems messy, seems like you could do the same with a preprocess hook to set the `$datalayer` variable (would have to be an object or array) and then have one line in twig to render it to JSON.

akf: Yes, I think that's a good idea.

omerida: i'll file a jira ticket

akf (changes requested):
Love the comma! 

I should have noticed this before -- this implementation puts out 

```"hasBenefitCategory": true,```
 
on pages that could be tagged, regardless of whether any categories are tagged. So I see this value as true on /flag, for example. Looking at the original request and also your description on the PR, we do want "false" in that case. (So it's a bit redundant wit the presence of the list.)

omerida:
Sorry. I should have caught that too after consolidating the logic. I've updated and checked that it outputs false on /flag, true + category list on a page that's been tagged.

I noticed this only adds these for "Standard Pages" but not on other node types like a Life Event in benefit finder. I'm wondering if that's OK ...

akf:
> I noticed this only adds these for "Standard Pages" but not on other node types like a Life Event in benefit finder. I'm wondering if that's OK ...

I'm fairly certain we want it on all pages. We need it in that "else" clause starting on line 100 of the html.html.twig file.

The Analytics team could probably work with not having it on all pages, but I think it's easier if they can just rely on having it everywhere.

omerida:
> I'm fairly certain we want it on all pages. We need it in that "else" clause starting on line 100 of the html.html.twig file.

All set and tested

akf (approved):
Checked a variety of content types and page types; looks good!

PR #1754, opened 2024-07-25
USAGOV-1727 - /node/ paths in Published Pages Report (static site generator)

DaleMHFrey:
https://cm-jira.usa.gov/browse/USAGOV-1727

## Description
* Fixed a situation where trying to change an alias of a Spanish Basic-Page node would crash the PHP-thread with an error of; `External URLs do not have internal route parameters `
* Fixed a situation where the taxonomy/breadcrumbs (found in the page source-code around the `dataLayer` area) were not always loading the alias-paths for certain nodes
    * this seemed to be an issue with Basic-Page nodes set to the Spanish language

## Type of Changes

- [ ] New Feature
- [X] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

## Testing Instructions
1.Edit a node like `/node/1919/edit` and try to change its alias-path. Notice how it crashes on Dev/Test. Now try again after this merge.
1. Run the static-site generator. Before this merge `/node/1919` shows up in the static-site generator. After this marge, its alias path is used instead.

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

4. Find the commit of this pull request.
5. Build and deploy the changes.
6. Update the Jira ticket by changing the ticket status to `Review in Test` and add a comment. State whether the change is already visible on [cms-dev.usa.gov](http://cms-dev.usa.gov/) and [beta-dev.usa.gov](http://beta-dev.usa.gov/), or if the deployment is still in process.


web/themes/custom/usagov/usagov.theme
```diff
@@ -10,6 +10,32 @@ use Drupal\taxonomy\Entity\Term;
...
+ * so in here we will notice when we are on a /node/#### page, and manually lookup the alias
+ * from the path_alias table.
+ */
+function usagov_preprocess_block__usagov_taxonomy_links(&$variables) {
+
+    // In theory, this should return an alias-path, but in certain edge-cases, we get the system path.
+    $urlToPass = \Drupal::request()->getRequestUri();
+
+    // If we are NOT seeing an alias path...
+    if ( substr($urlToPass, 0, 6) === '/node/' ) {
```
akf, line 24: If we go forward with this, let's log a warning here, so we can look for this kind of thing in our logs and know when we have problems to fix.

akf (changes requested):
This led me to look at the path aliases in a more specific way, and I think I found the cause of these particular paths getting picked up in static site generation. I put a lengthy explanation on the Jira ticket. 

I'm flagging this as "changes requested," but I'm not sure whether we want to keep any of this approach. What I'm seeing is that the dataLayer values are closer to correct in the /node/ directory with this update, but they're not really right (the paths for Spanish-language pages should start with /es) and they do already come out correctly when the page is accessed from the "correct" path alias.

I do like the idea of inserting a warning when an alias path isn't found.

web/themes/custom/usagov/usagov.theme
DaleMHFrey: Sure, just added the warning in (on now-line 27). Not sure if this matters, but should I not label this as a warning from "usagov_ssg_postprocessing"? The line really comes from this theme file, but I know the SSG will be the thing to really pop out the warnings when they do hit

akf:
After much discussion, we've decided we just needed to fix one piece of data and we don't need to make code changes.

PR #1758, opened 2024-07-25
USAGOV-1753 - Alpine-based generic cron app

arpage:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1753

## Description

Create the cron app

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [x] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [x] Cron
  - [ ] Tools
- [ ] Other

### Change Requirements

- [x] Requires New Documentation (Link: {})
- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [x] Communication with external service
- [x] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


.docker/src-cron/etc/ssl/certs/nginx-selfsigned.crt
akf, line 1: Does this app use nginx? Does it use rds (for which I also see certs)?

arpage: removed nginx and rds cert files

.circleci/config.yml
akf, line 1: CircleCI is not happy with this config: 
```Error calling workflow: 'build-and-deploy-cron'
Cannot find a definition for job named deploy-cron-to-env```

We should also have scan-code and scan-dependencies similar to what we do when building the cms, waf, and www containers. 

If I recall correctly, you were going to deploy this into each (dev, stage, prod) space, so perhaps it makes sense to bundle that together in one big honking deployment pipeline.

File: .docker/Dockerfile-cron [hunk 1]
```diff
@@ -0,0 +1,85 @@
...
+ENV CONTAINERTAG ${CONTAINERTAG:-none}
+
+ARG BUID=1000
+ARG BGID=1000
+
+# Dependencies
+RUN sed -E -i "s/:x:${BUID}:/:x:1919:/g" /etc/passwd \
+    && sed -E -i "s/:x:([0-9]+):${BGID}:/:x:\1:1919:/g" /etc/passwd \
+    && sed -E -i "s/:x:${BGID}:/:x:1919:/g" /etc/group \
+    && sed -E -i "s/nginx:x:[0-9]+:[0-9]+:/nginx:x:${BUID}:${BGID}:/g" /etc/passwd \
```
akf: Do we need an nginx user? I don't see any processes owned by nginx on the current app. (Also that would be kind of weird since nginx isn't installed.)

.docker/src-cron/opt/cron/bootstrap.sh
akf, line 1: Let's add `echo "Deployment: bootstrap complete"` at the end of this. Right now /tmp/bootstrap.log just has this one line about bootstrap starting, which kind of leaves me hanging.

.docker/src-cron/etc/ssl/certs/dhparam.pem
akf, line 1: I don't think we need this. In other apps it's specifically called out as part of nginx config, so I'm fairly confident it's just something nginx needed.

.docker/src-cron/opt/cfevents/capture-latest-events
akf, line 1: I haven't gone through everything, but it looks like we're using our existing "storage" bucket for everything. I believe we should have at least two additional buckets for use by this app (buckets are practically free). 

* For the cron app and/or cfevents: a private bucket ("basic" plan). I think nothing is being saved there for public consumption, right? 
* For the call center wait times: a public bucket ("basic-public" plan), on which we will need to set a CORS policy so that a javascript call from a page served from usa.gov can retrieve an object. 

I'm flagging a message in Slack with notes on setting the CORS policy. I tested it out awhile ago and it works well; we might want to take another look at the AllowedOrigins and add our "cms" urls and the ones from the dr space. I think I captured exactly what I did for a put-bucket-cors command, but I'm not sure where. It wasn't complicated.

.docker/src-cron/opt/cron/lock-singleton-task
akf, line 1: We talked about this before and this is smart, but a recent conversation reminded me of flock() -- maybe that would make this simpler? This would work for scripts that are running on the cron app but not for cf tasks they launch. Are we still looking at running cf tasks? I've lost the plot on that. 

I think flock() would take care of tracking the pid for you and would (IIRC, please verify) automatically release the lock when the process exits, whether cleanly or not. 

So it might go like: 

1. get my filename
2. flock(my_filename, LOCK_NB) # Non-blocking lock; otherwise we wait
3. if flock returned 0, we didn't get a lock so exit. (maybe log a warning)
4. otherwise proceed

web/themes/custom/usagov/scripts/callCenterWaitTimeLookup.js
```diff
@@ -0,0 +1,85 @@
+jQuery(document).ready(async function () {
+  "use strict";
+
+  var waittime = (function() {
+    if (jQuery("html[lang|='en']").length ||
+        jQuery("html[lang|='es']").length) {
+      var domain = window.location.origin.split('.')[0].split('//')[1];
+      var domainMap = {
```
akf, line 8: I have really been trying to get us away from this! We should not be calling the CMS for this information. 

Instead, we should be making an https request of an s3 bucket (like the analytics dashboard's javascript does). 

There's also an implication here that we really should be running the call center cron jobs only on one system, since they will write data to a bucket everyone can share. It's fine for the code to be in all three environments and we should have a way to turn it on or off so we'll be able to test changes going forward. That could be as simple as a 1/0 environment variable on the app.

web/themes/custom/usagov/scripts/callCenterWaitTimeLookup.js
```diff
@@ -0,0 +1,85 @@
...
+          var json = jQuery.parseJSON(response);
+          var seconds = -1;
+          if (jQuery("html[lang|='en']").length) {
+            seconds = json.enEstimatedWaitTimeSeconds;
+          }
+          if (jQuery("html[lang|='es']").length) {
+            seconds = json.spEstimatedWaitTimeSeconds;
+          }
+
+          if (seconds >= 0) {
```
akf, line 31: I would also like the ccwaittimes script to include a timestamp in the file it writes to s3, and for this script to check it, so that if the former has stopped updating the file, the latter will detect that and not display an estimate. 

(I realize these comments are really for Jacob.)

scripts/cron/full-rebuild.sh
akf, line 1: These (rebuild.sh and full-rebuild.sh) look reasonable enough, but let's make sure we document them and when to use them. (Presumably in the normal case people will just rely on what's hooked into CI?) 

That might just mean "put a comment at the top" but I'm thinking maybe a readme about the cron app is in order.

akf: Same goes for the cfevents versions of these.

scripts/cfevents/capture.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+SPACE=$1
+
+aws s3 rm --recursive s3://$S3_BUCKET/cfevents/$SPACE/
+
+cf run-task cfevents --name cfevents-instance --command "/opt/cfevents/capture-latest-events $SPACE"
+
+sleep 15
```
akf: I don't  like this timing assumption, but also I think we'll be able to drop this once cfevents is emitting log messages and those are flowing to NR and the log-shipper's s3 bucket. Let's make some time to talk about this.

akf (commented):
There's a bunch of this I haven't reviewed yet, but I'm giving you the comments I have so far.

.docker/src-cron/opt/callcenter/call-center-update
akf, line 1: I think I wrote this elsewhere, but I would like this script to add a timestamp to the waittime.json file. The client-side script callCenterWaitTimeLookup.js would then check the timestamp and treat the data as "missing" if it's sufficiently out-of-date. (Since we're proposing to run this every minute, "sufficiently out of date" could be 5 minutes.)

.docker/src-cron/root/.profile
akf, line 1: Since I'm determined to have multiple s3 buckets, this script will need some changes. Here's one approach that might make sense: 

- Each bucket should have a sensible name. For the sake of argument, let's say "private-storage" and "public-with-cors". 
- Any s3 bucket that's needed by a cron script (and that runs directly in the cron app) gets bound to the cron app, so we can retrieve credentials much as we do here. (Yes, we can bind a bucket to apps in more than one space!) 
- The aws_cp, aws_rm, and aws_ls commands would take the cf bucket name (e.g., "public-with-cors") and set AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_DEFAULT_REGION for that particular bucket. 
- (I don't know whether it makes more sense to gather those once and keep a data structure locally, or just pull them from $VCAP_SERVICES on the fly. I think probably $VCAP_SERVICES every time is better.) 
- We should be able to do this without explicitly calling out the bucket names in this script, I think. We just need to bind the buckets to this app.

.github/pull_request_template.md
akf, line 1: 👍

bin/cloudgov/container-build-callcenter
akf, line 1: Remove (right? My understanding is we're not launching a "callcenter" container, and I don't see Dockerfile-callcenter on this branch)

bin/cloudgov/container-push-callcenter
akf, line 1: Also remove -- no callcenter container

bin/cloudgov/deploy-callcenter
akf, line 1: I think we can get rid of this script as well, but that raises the question -- how is the call-center-update script getting these environment variables?

scripts/cfevents/README-CF-TASKS.md
akf, line 1: Consider moving this into the top-level /docs directory. It might need updates, but I'm not going into that right now. 
Is there a README for adding this directly to the cron app as well? (I think not, but I might be missing something. Probably it should cover passing in credentials. Currently my thought is that we should add these as user-provided services, rather than $ENV, mostly because it seems easier to manage them on a per-task basis that way.)

.docker/src-cron/opt/callcenter/env.default
akf, line 1: This isn't used anywhere anymore, is it? It's basically just serving as a reminder that we need to provide these?

.docker/src-cron/opt/cfevents/app-status
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
```
akf, line 2: I do like to have a comment at the top of a script that says what it's for. Is this just meant to be run for debugging from the command line? (If not, I have concerns about some of what it's echoing.)

.docker/src-cron/opt/cron/app-status
akf, line 1: Same connets as on the cfevents/app-status script -- let's add a comment about what it's for, and enlighten me.

.docker/src-cron/opt/cron/cert-watcher.sh
```diff
@@ -0,0 +1,47 @@
...
+
+    # copy cloud foundry certificates
+    if [ -d "$CF_SYSTEM_CERT_PATH" ]; then
+      cp $CF_SYSTEM_CERT_PATH/*  /usr/local/share/ca-certificates/
+    fi
+
+    # load these certs
+    /usr/sbin/update-ca-certificates 2>&1 > /dev/null || echo ""
+
+    # Kill the newrelic-daemon. It will restart automatically and read the certificates.
```
akf, line 40: Remove commented-out code. (I'm not just being pedantic, it's good for future greppers.)

.docker/src-cron/opt/cron/cf-task-check
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+
```
akf, line 2: Add a comment about what this is for, please. 
I don't see anything calling this; is it leftover? Or a utility script for developers?

bin/cloudgov/deploy-cfevents
akf, line 1: It looks like we're not using this script (yet?) so I'm just giving it a light review.

bin/cloudgov/deploy-cron
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
+
```
akf, line 2: I should ask for a comment on this one too, but our comments on the deploy-APPNAME scripts are so vacuous ...

bin/cloudgov/deploy-cron
```diff
@@ -0,0 +1,108 @@
...
+SPACE=$(echo "$SPACE" | tr '[:upper:]' '[:lower:]')
+assertCurSpace $SPACE
+shift
+ORG=$(getOrg)
+
+DOCKERUSER=${DOCKERUSER:-gsatts}
+DOCKERREPO=${DOCKERREPO:-usagov-2021}
+
+export APPNAME="cron"
+APPSPEC=${DOCKERUSER}/${DOCKERREPO}:${APPNAME}
```
akf: Unless I'm missing something, this script is always to push the tag "cron" (and apparently we're always saving the image as simply "cron," not "cron-latest")? We need to maintain the pattern established for the cms/www/waf containers, where we store the image digest when we push to docker hub, and use the image digest to specify the image later. 

We might as well also maintain the pattern of pushing that image as "appname-<pipeline-number>" and as "appname-latest" as well. This is less critical, but it is really nice to be able to visually match up the images in docker hub with CI pipeline numbers if something doesn't seem right. 

(The reason for the digests was that cf simply doesn't support "docker content trust," so we can't rely on that.)

bin/cloudgov/events/get-events
```diff
@@ -1,6 +1,4 @@
-#!/bin/env bash
-
-source ~/.profile &> /dev/null
+#!/usr/bin/env bash
 
 # we might be running in circleci
```
akf, line 8: Please include a comment about what this script is for.

bin/cloudgov/deploy-cron
akf: I also want to acknowledge that this is set up to only deploy a single instance, with no option to change that. I think that's probably the right thing for this app, in spite of its "production" status. Probably there should be a comment near the "cf push" command to that effect.

bin/cloudgov/events/get-events
akf, line 1: I notice that this script does echo the event output (csv) to STDOUT as well as tee'ing it to a file. So I made a change to someone's roles in the dr space (and then changed it back), and I see that New Relic now has the logs. They are, of course, showing up as a comma-separated string in the "message" field.

akf (changes requested):
Here are a bunch more comments. 

I think I pointed out earlier that CircleCI isn't happy with the deployment script right now. I've been supplementing my review by looking at what's happening with the "cron" app that's currently up and running, but it's probably a bit out of date from this code. 

There's a lot here! Thanks for all your work on this so far!

.docker/src-cron/opt/cron/lock-singleton-task
arpage: Flock would indeed be simpler.  The only thing it would lack (which the current implementation already has a hook for [see line 46 of lock-simpleton-task]) is a means to intervene if a process is running long or hung up.

Whether that is worth the extra complexity is a topic worth discussing further :)

akf: _lock-simpleton-task_ 😄

arpage: o.m.g.  This is what happens when I write comments late at night :)

What do you think about the topic though?

.docker/src-cron/opt/cron/cf-task-check
arpage: This is the beginnings of the serializing lock implementation for cf task based jobs.

.docker/src-cron/opt/cron/app-status
arpage: This is mainly for task-based apps (ported over from cfevents/callcenter apps) to provide the equivalent of "drush-status", since we cannot ssh into the task and query it's environment.

.docker/src-cron/opt/cfevents/app-status
arpage: Equivalent of "drush-status" for a cf task based app.  See comment below.

.docker/src-cron/opt/callcenter/env.default
arpage: This is correct.   Can be removed if desired.

bin/cloudgov/deploy-cron
arpage: Fixed the build/push/deploy scripts to behave like the rest of our apps:  handle tag, or digest if provided.   Parameterized and commented the instance count a bit, but otherwise it's still a single-instance app by default.

bin/cloudgov/deploy-callcenter
arpage: @akf - these are in env.local, and should also be placed in CircleCI.

.docker/src-cron/opt/cfevents/capture-latest-events
arpage: Created separate buckets, and added CORS policy to callwaittime bucket.

.docker/src-cron/root/.profile
arpage: Each "thing" can have it's own service/bucket.  Providing a second argument to root/.profile will select the service/bucket to use. with no argument or an unknown service/bucket defaulting to the cron app's bucket.

arpage: Also, CORS policy is addressed in the deploy-cron script, where the services/buckets are created if needed.

web/themes/custom/usagov/scripts/callCenterWaitTimeLookup.js
arpage: See USAGOV-1945 - moving this process to that ticket

.docker/src-cron/opt/callcenter/call-center-update
arpage: See USAGOV-1945 - moving this process to that ticket

web/themes/custom/usagov/scripts/callCenterWaitTimeLookup.js
arpage: See ticket USAGOV-1945

.docker/src-cron/opt/cfevents/app-status
akf: I'm trying to put a finger on why the comment below didn't do it for me. I think it's the wording "For task-based apps, provide a means of querying the run-time environment of the task." Just changing "provide" to "provides" would help a little. But it could be more direct. How about: 

```
# Run this script as a task to query the run-time environment. 
# e.g., cf run-task my-app "/opt/cfevents/app-status"
```

I realize this is small potatoes.

.docker/src-cron/root/.profile
```diff
@@ -0,0 +1,126 @@
...
+      cp $CF_SYSTEM_CERT_PATH/*  /usr/local/share/ca-certificates/
+   fi
+   /usr/sbin/update-ca-certificates
+   touch ~/.certs-updated
+fi
+
+CF_USERNAME=$(echo "$VCAP_SERVICES" | jq -r '.["cloud-gov-service-account"][]? | select(.name == "cron-service-account") | .credentials.username';)
+CF_PASSWORD=$(echo "$VCAP_SERVICES" | jq -r '.["cloud-gov-service-account"][]? | select(.name == "cron-service-account") | .credentials.password')
+
+case $TASK in
```
akf, line 47: nice!

.docker/src-cron/root/.profile
```diff
@@ -0,0 +1,126 @@
...
+   *)
+      export S3_BUCKET=$(echo "$VCAP_SERVICES"             | jq -r '.["s3"][]? | select(.name == "cron-state-storage") | .credentials.bucket')
+      export S3_ENDPOINT=$(echo "$VCAP_SERVICES"           | jq -r '.["s3"][]? | select(.name == "cron-state-storage") | .credentials.fips_endpoint')
+      export AWS_ACCESS_KEY_ID=$(echo "$VCAP_SERVICES"     | jq -r '.["s3"][]? | select(.name == "cron-state-storage") | .credentials.access_key_id')
+      export AWS_SECRET_ACCESS_KEY=$(echo "$VCAP_SERVICES" | jq -r '.["s3"][]? | select(.name == "cron-state-storage") | .credentials.secret_access_key')
+      export AWS_DEFAULT_REGION=$(echo "$VCAP_SERVICES"    | jq -r '.["s3"][]? | select(.name == "cron-state-storage") | .credentials.region')
+      ;;
+esac
+
+echo S3_BUCKET: $S3_BUCKET
```
akf: We'll want to remove these echo statements before going live (and re-create the buckets in the dr space, I think). I'm surprised not to see this in the logs, so maybe it's not exposing the secrets. (I do see that you usually redirect errors to /dev/null when running this script from another.)

bin/cloudgov/s3-clear-bucket
akf, line 1: I think I spy missed items here, and possibly also a shorter solution. 
Missed: objects at top-level (not in a "folder.")
Simpler: `aws s3 rm --only-show-errors --recursive s3://$S3_BUCKET/ `

Honestly I haven't tested this, but two sources I found on the internet say ti should work.

s3-cors-policy-callwaittime.json
```diff
@@ -0,0 +1,27 @@
...
+            "AllowedOrigins": [
+                "https://cms-dr.usa.gov",
+                "https://cms-dev.usa.gov",
+                "https://cms-stage.usa.gov",
+                "https://beta-dr.usa.gov",
+                "https://beta-dev.usa.gov",
+                "https://beta-stage.usa.gov",
+                "https://cms.usa.gov",
+                "https://usa.gov",
+                "https://www.usa.gov"
```
akf: Oh hey, let's add '"http://localhost"'
It will help people test locally!

web/themes/custom/usagov/usagov.libraries.yml
```diff
@@ -8,6 +8,7 @@ global-styling:
     scripts/uswds-init.min.js: {}
     scripts/newFooter.js: {}
     scripts/phoneForMobile.js: {}
+    scripts/callCenterWaitTimeLookup.js: {}
```
akf, line 11: Let's comment this out for now.

akf (commented):
I don't think you were ready for me, but I gave this another run-through so I can be speedier when you are ready. Just added some comments this time.

.docker/src-cron/root/.profile
arpage: Removed echo statements and redirected cf's verbose commands.  Captured return values, so we can still notify on error.

.circleci/config.yml
```diff
@@ -727,3 +824,92 @@ workflows:
...
+            - approve-build-and-push-container-cron
+          filters:
+            branches:
+              only:
+                - dr
+                - dev
+                - stage
+                - prod
+                - USAGOV-1753-generic-cron-app-alpine
+      - approve-dev-deployment-cron:
```
akf, line 851: My recollection is that we need one of these in each space, right? Or will the one in `prod` be able to log events for all spaces? 

If we need one for each space, we definitely need one in `shared-egress` and `tools`. I'd be happy to have that be a follow-on to this effort; just wanted to make note of it.

.docker/src-cron/root/.profile
```diff
@@ -0,0 +1,139 @@
...
+
+if [ ! -f ~/.certs-updated ]; then
+   echo "... Combining and updating certs ..."
+   if [ -d "$CF_SYSTEM_CERT_PATH" ]; then
+      cp $CF_SYSTEM_CERT_PATH/*  /usr/local/share/ca-certificates/
+   fi
+   /usr/sbin/update-ca-certificates
+   touch ~/.certs-updated
+fi
+
```
akf, line 43: Somewhere around here is where I think you should set `HTTPS_PROXY` et al to `$PROXYROUTE`. See my corresponding note on `deploy-cron`.

bin/cloudgov/deploy-cron
```diff
@@ -0,0 +1,147 @@
...
+    if [ "$proxySet" = "1" ]; then
+        echo "Setting http_proxy to $proxyName"
+        cf target -s $proxySpace -o $proxyOrg > /dev/null 2>&1;
+        proxyroute=$(cf env $proxyName | grep proxy_route | awk '{print $2}')
+        if [ -z "$proxyroute" ]; then
+            echo "ERROR: $proxyName does not have a proxy_route set"
+            exit 1
+        fi;
+        cf target -s $appSpace -o $appOrg > /dev/null 2>&1
+        cf set-env $appName https_proxy $proxyroute
```
akf: We're going to regret doing `cf set-env $appName https_proxy` here. We've adopted (maybe just I've adopted, but it's a thing we're doing) this convention: 

> Use `cf set-env` to set PROXYROUTE on the app, then if the proxy is actually desired, set HTTPS_PROXY and any similar variables you need within the app's startup scripts. 

This does a couple of things for us: 
- It lets us be a bit more selective about when we use the proxy. 
- It means that we don't invoke the proxy during staging, which is a problem for some apps. (I think this was an issue.)

The reason this will cause us pain in the future is that the setup-proxy-for-apps script updates PROXYROUTE only. So if we update the proxy and rotate its credentials, they won't get copied to HTTP[S]_PROXY and http[s]_proxy on this app. 

So, in the case of this app, I think we should set HTTP[S]_PROXY and http[s]_proxy in root/.profile.

bin/cloudgov/setup-egress-for-apps
```diff
@@ -145,9 +147,22 @@ for app in $applist; do
     $output cf set-env "$app" AWS_CA_BUNDLE /etc/ssl/certs/ca-certificates.crt
     $output cf set-env "$app" REQUESTS_CA_BUNDLE /etc/ssl/certs/ca-certificates.crt
 
+    ### Ensure that our primary apps are all attached to the log drain service!
```
akf, line 150: We need to ensure that the log-shipper-drain service does _not_ get bound to the log-shipper itself! (So far we seem to have avoided this on the log-shipper in the dr space.) 

There's a little script in the usagov-logshipper repo to do this: bin/add-log-drains-for-space.sh. And there is a circleci action for it in that repo, but of course no one finds it. So I don't object to adding it here, but we should include that clause: 
```    if [[ ! "$app" = "log-shipper" ]]; then``` 

(I've come to believe that we should fold the logshipper code back into the usagov-2021 repo, sigh.)

File: scripts/cron/rebuild.sh
Comment (akf) [inline scripts/cron/rebuild.sh:1]: Is this something that should be added to the "building USAGov from scratch" instructions?

**Review by akf — CHANGES_REQUESTED** [review]
It seems I still have a couple of substantive, let's-do-this-before-merging, change requests related to the log drain and proxy: 

* Set HTTPS_PROXY and friends within the app (probably in .profile), not with `cf set-env`. That's because the proxy-related deploy scripts expect to maintain the PROXYROUTE variable only. If they're used to update it, an app that relies on `cf set-env` for the other variables will miss it. 
* Ensure that log-shipper-drain won't be bound to the log-shipper.

I think my other comments are things that can wait; I might have forgotten something.

File: bin/cloudgov/s3-clear-bucket
Comment (arpage) [reply bin/cloudgov/s3-clear-bucket:1]: this works. adding - ty!

File: web/themes/custom/usagov/usagov.libraries.yml (hunk 14 above)
Comment (arpage) [reply web/themes/custom/usagov/usagov.libraries.yml:11]: Moving this to USAGOV-1945 with the other call center changes.

File: .circleci/config.yml (hunk 15 above)
Comment (arpage) [reply .circleci/config.yml:851]: USAGOV-1962 created and added to Logging and Compliance epic.

File: bin/cloudgov/setup-egress-for-apps (hunk 18 above)
Comment (arpage) [reply bin/cloudgov/setup-egress-for-apps:150]: OK, adding conditional here.

File: scripts/cron/rebuild.sh
Comment (arpage) [reply scripts/cron/rebuild.sh:1]: It is, and has been added.

File: .docker/src-cron/root/.profile (hunk 16 above)
Comment (arpage) [reply .docker/src-cron/root/.profile:43]: Adding the env vars closer to the beginning, in case something else needs them.

File: bin/cloudgov/deploy-cron (hunk 17 above)
Comment (arpage) [reply bin/cloudgov/deploy-cron]: OK, changed to only setting PROXYROUTE via `cf set-env` and then setting HTTP_PROXY and friends via `~/.profile`

File: .docker/src-cron/opt/cfevents/app-status [hunk 19]
```diff
@@ -0,0 +1,50 @@
...
+# Run this script as a task to query the run-time environment.
+# e.g., cf run-task my-app "/opt/cfevents/app-status"
```
akf: Okay, as long as I'm here, I tried this and discovered that's a syntax error. It needs `-c` or `--command` before the command string.

.docker/src-cron/opt/cron/app-status
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+
+# For task-based apps, provide a means of querying the run-time environment of the task
+# e.g.  cf run-task my-app "/opt/myapp/app-status"
```
akf: add `-c` or `--command`. 
Also I'm thinking we should omit the proxy variables from the output, as they include (admittedly low-value) credentials. 

It would be nice (in the future) to instead indicate whether the proxy is set, and if it is, the result of executing curl against the expected external resource.

akf (commented):
I added a couple of "nice to do" comments.

.docker/src-cron/opt/cron/app-status
arpage: Implemented.

.docker/src-cron/opt/cfevents/app-status
arpage: Implemented

akf (approved):

PR #1772, opened 2024-07-30
USAGOV-1789-wizard-datalayer: Send required information to the datalayer on wizard taxonomy pages. 

mdranove:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1789

## Description

Use theme suggestions alter hook in usagov_wizard.module to add new html twig template for wizard pages. Use preprocess hook to send information to new twig template. Have MenuChecker.php return a new variable which is the entity object of the wizard entity that is linked to from the left hand navigation menu.

## Type of Changes

- [ X ] New Feature
- 
image

- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

## Testing Instructions

Repeat all the steps with a few different pages.

#### Taxonomy ID
1. Visit a wizard taxonomy page.
2. Discover the entity id by hovering over the "edit" button looking at the bottom left of your screen where the link is previewed, and noting the number that appears in "/taxonomy/term/[number]". 
3. Open devtools and ctrl+f "dataLayer" in the DOM. 
4. Check the value for "taxonomyID". 
5. The value matches the number in step 2.

#### Content type, language, homepageTest, basicPageSubtype, hasBenefitCategory
1. Check the dataLayer "contentType" tag.
2. The values all match the ones that are in the dataLayer of the current scams wizard on production.

#### Page_Type
1. Navigate to the root of the wizard, the page with the title **Learn where to report a scam**.
2. Check the datalayer.
3. The value for the "Page_Type" tag is "wizard-start".
4. Navigate to the next page. 
5. The value for the "Page_Type" tag is "wizard-question".
6. Repeat step 4-5 with a couple of other pages.
7. Navigate to the last page in any part of the tree.
8. The value for the "Page_type" tag is "wizard-result"

#### Taxonomy Text and Taxonomy URL
This was the most complicated to implement, and most error prone, so probably if anything needs special testing attention it's this.
1. Go to the start of the wizard
2. Check the datalayer starting from the first Taxonomy_Text tag.
3. The Taxonomy_Text_3 tag matches the name of the menu link on the left hand navigation.
4. All the other Taxonomy_Text tags match those on production.
5. The Taxonomy_URL tags match the aliases given to each page listed.
6. Repeat with a number of different pages in different parts of the tree, keeping in mind that the taxonomy_text and taxonomy_url tags will start to look different than what is on production since the titles of the pages have changed. Just ensure that the text in these tags are expected. E.g, if the title of the page is "Test", then the final taxonomy_text tag value should be "Test". And if the page before "Test" in the tree is "Another Test", then the second to last taxonomy_text tag value should be "Another Test".

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

10. Find the commit of this pull request.
11. Build and deploy the changes.
12. Update the Jira ticket by changing the ticket status to `Review in Test` and add a comment. State whether the change is already visible on [cms-dev.usa.gov](http://cms-dev.usa.gov/) and [beta-dev.usa.gov](http://beta-dev.usa.gov/), or if the deployment is still in process.

### DIFF
--- a/web/themes/custom/usagov/templates/html.html.twig
+++ b/web/themes/custom/usagov/templates/html.html.twig
@@ -23,6 +23,7 @@
  * @see template_preprocess_html()
  */
 #}
+
 {% set lang = node.langcode.value %}
 {% set nodeID = node.nid[0].value %}
 {% set path = drupal_url('node/'~nodeID, {absolute: TRUE}) %}
@@ -38,20 +39,23 @@
 <!DOCTYPE html>
 <html{{ html_attributes }}>
 <head>
-{% if meta_noindex %}
-<meta name="robots" content="noindex" />
-{% endif %}
+{% block header_html %}
+  {% if meta_noindex %}
+  <meta name="robots" content="noindex" />
+  {% endif %}
 <link rel="preconnect" href="https://script.crazyegg.com" crossorigin>
 <link rel="preconnect" href="https://www.googletagmanager.com" crossorigin>
 <head-placeholder token="{{ placeholder_token|raw }}">
 <title>{{ head_title|safe_join(' | ') }}</title>
   <css-placeholder token="{{ placeholder_token|raw }}">
   <js-placeholder token="{{ placeholder_token|raw }}">
+
   {% if node.id is defined %}
     {{ node.field_header_html.value|raw }}
-  {% elseif header_html %}
-    {{ header_html|raw }}
   {% endif %}
+{% endblock %}
+
+{% block taxonomy_data %}
   {% if nodetype == "basic_page" %}
     <script id="taxonomy-data">
     dataLayer = [{
@@ -111,6 +115,7 @@
       }];
     </script>
  {% endif %}
+{% endblock %}
   <!-- Google Tag Manager -->
   <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
   new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
--- a/web/themes/custom/usagov/templates/wizard_2024/html--taxonomy--term--wizard.html.twig
+++ b/web/themes/custom/usagov/templates/wizard_2024/html--taxonomy--term--wizard.html.twig
@@ -1,0 +1,117 @@
+{% extends "html.html.twig" %}
+
+{% block header_html %}
+  {% if not start_page %}
+    <meta name="robots" content="noindex" />
+  {% endif %}
+    <link rel="preconnect" href="https://script.crazyegg.com" crossorigin>
+    <link rel="preconnect" href="https://www.googletagmanager.com" crossorigin>
+    <head-placeholder token="{{ placeholder_token|raw }}">
+      <title>{{ head_title|safe_join(' | ') }}</title>
+      <css-placeholder token="{{ placeholder_token|raw }}">
+        <js-placeholder token="{{ placeholder_token|raw }}">
+  {# tracking scripts, set in the field. #}
+  {% if header_html %}
+    {{ header_html|raw }}
+  {% endif %}
+{% endblock %}
+
+{% block taxonomy_data %}
+  {% set titleCounter = 4 %}
+  {% set urlCounter = 4 %}
+  {% if start_page %}
+    {% set page_type = 'wizard-start' %}
+    {% set titleCounter = 2 %}
+    {% set urlCounter = 2 %}
+  {% elseif not has_child %}
+    {% set page_type = 'wizard-result' %}
+  {% else %}
+    {% set page_type = 'wizard-question' %}
+  {% endif %}
+  <script id="taxonomy-data">
+    {# passing this info from the preprocess_hook doesn't work due to how the
+    page loads this template, so just load the breadcrumb block in the template,
+    and then use twig magic to get all the info we need. #}
+    {% set breadcrumb_block = drupal_block('system_breadcrumb_block') %}
+    {# this is hard to to explain, but basically since the breadcrumb block
+    doesn't give us a link to the current page, we need to use logic from the
+    preprocess hook to print the current page info. We also need to print
+    information about the basic page landing page this is linked to from, and
+    we need to number both entries in the datalayer correctly.
+
+    E.g all the pages should have 'Taxonomy_Text_1' be the homepage, and
+    'Taxonomy_Text_2' be 'scams and fraud', but only the first page in the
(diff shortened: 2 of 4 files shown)


akf (commented):
This looks good and I see one thing you'll need to modify: I know the Analytics team wants six entries for "Taxonomy_Text" and "Taxonomy_URL" even if (as is usual) there aren't six levels. They should just repeat. Here's an example from prod: 

```
"Taxonomy_Text_1": "Home",
 "Taxonomy_Text_2": "Scams and Fraud",
"Taxonomy_Text_3": "Where to report a scam",
"Taxonomy_Text_4": "What type of scam do you need to report?",
"Taxonomy_Text_5": "What type of scam do you need to report?",
"Taxonomy_Text_6": "What type of scam do you need to report?",
"Taxonomy_URL_1": "/",
"Taxonomy_URL_2": "/scams-and-fraud",
"Taxonomy_URL_3": "/where-report-scams",
"Taxonomy_URL_4": "/where-report-scams/what-type-scam-do-you-need-report",
"Taxonomy_URL_5": "/where-report-scams/what-type-scam-do-you-need-report",
"Taxonomy_URL_6": "/where-report-scams/what-type-scam-do-you-need-report"
```

mdranove:
> This looks good and I see one thing you'll need to modify: I know the Analytics team wants six entries for "Taxonomy_Text" and "Taxonomy_URL" even if (as is usual) there aren't six levels. They should just repeat.

Hi Amy, thanks for the comment. I've gone ahead and modified the logic in the twig template only. Can you please do a full retest of the testing steps from **Taxonomy Text and Taxonomy URL**? I reworked the logic quite a bit.

akf (changes requested):
I found two "real" things and one "maybe": 

**Critical**: Non-wizard pages have some new datalayer problems: 

* On the home pages ( / and /es ), `nodeID` and `location` are both null, and the `Taxonomy_*` entries are entirely missing. 
* On /passport and /apply-adult-passport (as examples of other page types), `nodeID` and `language` are null as well. The `Taxonomy_*` entries do appear correctly, though. 

I don't think I checked for this during my previous review, so this might not be new with the latest refactor. I did find that I could "fix" this by removing the `{% block variables %}` markup from html.html.twig. 

**Easy**: The last entry in the dataLayer object shouldn't have a comma after it. Honestly I don't know whether there are any tools that are going to care about that. 

**Maybe not a problem**: Some of my wizard-result pages don't have path aliases, and I see mismatched IDs, for example: 

```"taxonomyID": "120",```
```"Taxonomy_URL_6": "/taxonomy/term/103",```

FYI, @omerida  has embarked on a refactor of the dataLayer code overall. I don't know whether the timing works out so that might save you any grief, but I'll flag you on the corresponding ticket.

**Comment by omerida** [comment]
Here's a draft PR with what I have working so far. I need to test more pages against it, to make sure the datalayer output is the same in all cases, so it's a proof-of-concept at the moment. https://github.com/usagov/usagov-2021/pull/1791/files

And I need to document it formally but the wizard module could modify the datalayer object with an event subscriber. Check out the EventSubscriber in usagov_benefit_category_search module.

**Comment by mdranove** [comment]
> Here's a draft PR with what I have working so far. I need to test more pages against it, to make sure the datalayer output is the same in all cases, so it's a proof-of-concept at the moment. https://github.com/usagov/usagov-2021/pull/1791/files
> 
> And I need to document it formally but the wizard module could modify the datalayer object with an event subscriber. Check out the EventSubscriber in usagov_benefit_category_search module.

@omerida, that looks cool, I think that could work, but I will say that since it's still just PoC Ill stick with my approach for now. But I'm going to merge this into my local and fiddle with it. My approach seems somewhat hacky so doing this all in an event subscriber/hook seems cleaner.

@akf Weird issue. You're right that removing the twig block fixes the issues, I did some digging and have no clue why that should matter. Anyway, I removed the twig block, and also the last comma in the datalayer object.

As for the other comments, I don't think there's an issue. The mismatched id you pointed out was for the last page in the tree. I've confirmed with David Kaufmann that if the tree goes beyond the 6th level of depth info should stop getting passed to the datalayer. So basically you were on taxonomy/term/120, but the last tracked url in the tree was /taxonomy/term/103 because it had gone too deep at this point to keep track of. As for the missing aliases, that's just because no alias has been entered via content for the entity.

File: web/themes/custom/usagov/templates/html.html.twig [hunk 1]
```diff
@@ -276,7 +281,7 @@ const scrollFunc = () => {
   } else {
     scrollToTopButton.className = "";
   }
-};
+%};
```
akf: Typo here -- % doesn't belong. 
This actually is causing a syntax error and causing the "Top" button not to appear when I scroll down the home page

web/themes/custom/usagov/templates/wizard_2024/html--taxonomy--term--wizard.html.twig
```diff
@@ -0,0 +1,117 @@
...
+{% endfor %}
+{% for i in range(1, 6) %}
+{% if urlCounter <= 6 %}
+{% set lastUrlCounter = 'Taxonomy_URL_' ~ urlCounter %}
+  {{ lastUrlCounter|json_encode(constant('JSON_PRETTY_PRINT'))|raw }}: {{ currentPath|json_encode(constant('JSON_UNESCAPED_SLASHES'))|raw }}{% if urlCounter < 6 %},
+  {% endif %}
+{% endif %}
+{% set urlCounter = urlCounter + 1 %}
+{% endfor %}
+    %}];
```
akf: Here's another % that doesn't belong. It's causing output like this: 
```
"Taxonomy_URL_6": "/scam-reporting"    %}];
```

akf (changes requested):
Datalayer values look good to me. 

I see two stray % characters messing up javascript that's embedded in twig.
If you update just those I can sign off pretty fast.

web/themes/custom/usagov/templates/wizard_2024/html--taxonomy--term--wizard.html.twig
mdranove: Hi @akf, thank you for catching these. Need to look into how that happened. I've pushed up the changes.

akf (approved):
Looks good!

PR #1800, opened 2024-08-06
USAGOV-1761-cf-components-audit: Created cf components audit script

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1761

## Description
Script for auditing components in cf environment.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [X] Other

## Testing Instructions
Run bin/cloudgov/audit/cf-components

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


akf (changes requested):
This is looking promising and I've learned some things, like the fact that back in 2019 someone created a my.rubytest.gov domain in the sandbox-gsa org. I've also realized that I really should review this kind of script carefully before just running it! (Nothing bad happened, but if you were a bad actor it could have.) 

This is slow, and after watching several screens of `vote-SPACE-upkeep` tasks list out in my terminal, I think we need to make some changes: 

1. Don't list the tasks, just show the count of tasks and move on. I see that the `tasks` endpoint provides pagination with a `total_results` value. 
2. Limit everything to the gsa-tts-usagov org. Maybe take the organization as an optional argument (and if it's not provided, do all orgs). 
3. Similarly, provide a way to specify which spaces to list. 
4. If listing all orgs, DO NOT list out the users. I'm getting all the users in the sandbox-gsa org. I think that's basically every engineer doing work for TTS and using cloud.gov. Or actually, no ... I'm just getting the first page of users
5. Skip the buildpacks you're getting under "misc info." I got a list of several buildpacks we don't use, and it's missing some that we do (apt_buildpack, for example). I don't know if this is just standard buildpacks or what. (The buildpacks listed for specific apps look correct, though.) 

And, I'm sorry, it looks like this needs to account for pagination.

jacobaaronyeager:
> This is looking promising and I've learned some things, like the fact that back in 2019 someone created a my.rubytest.gov domain in the sandbox-gsa org. I've also realized that I really should review this kind of script carefully before just running it! (Nothing bad happened, but if you were a bad actor it could have.)
> 
> This is slow, and after watching several screens of `vote-SPACE-upkeep` tasks list out in my terminal, I think we need to make some changes:
> 
> 1. Don't list the tasks, just show the count of tasks and move on. I see that the `tasks` endpoint provides pagination with a `total_results` value.
> 2. Limit everything to the gsa-tts-usagov org. Maybe take the organization as an optional argument (and if it's not provided, do all orgs).
> 3. Similarly, provide a way to specify which spaces to list.
> 4. If listing all orgs, DO NOT list out the users. I'm getting all the users in the sandbox-gsa org. I think that's basically every engineer doing work for TTS and using cloud.gov. Or actually, no ... I'm just getting the first page of users
> 5. Skip the buildpacks you're getting under "misc info." I got a list of several buildpacks we don't use, and it's missing some that we do (apt_buildpack, for example). I don't know if this is just standard buildpacks or what. (The buildpacks listed for specific apps look correct, though.)
> 
> And, I'm sorry, it looks like this needs to account for pagination.

These have been addressed.

akf (commented):
I'm giving this another look. I see a couple of error messages; here they are with a line of context above and below: 

```
     gsa.benefits.gov (Last updated: 2024-08-23T18:08:43Z | Internal: false | Supported Protocols: ["http"])
curl: (3) URL rejected: No host part in the URL
bin/cloudgov/audit/cf-components: line 96: [: -gt: unary operator expected
   Space: stage (Last updated: 2024-08-08T20:48:19Z)
```

So this is telling me the second error message occurs on line 96. Probably the other error is from the command on line 95 (which I notice doesn't have `$()` around the cf curl command).

akf (changes requested):
In addition to that error message I commented on earlier, I'm unfortunately seeing that this report is missing a lot of information. 

What I did was run: 
```
bin/cloudgov/audit/cf-components > cf-components.txt
```
I didn't supply any arguments, although I see that I could. I'm pleased that it limited its queries to the gsa-tts-usagov org. 

I didn't dig into the code because I'm short on time; I'm just critiquing the output. 

The output is in https://docs.google.com/document/d/1RyKRFbsvoeTKvM2frwYjSNzHzi8SycO35DjBNtTrpnU/edit and I've commented on specific parts, but the main themes are: 

- many of the lists are missing information (e.g., we have more services than show up). I saw this specifically for: 
  - domains
  - services (in one case I didn't get any services even though there are several I see when I use `cf services` to list them)
  - routes
- The **Security Groups** listing tells me the rules for the security groups that are available, but not where we're using them -- which is what we really need. Take a look at what "cf security-groups" shows -- you get a table of name, organization, space, and lifecycle (running or staging). It makes sense for us to look for things like "public_networks_egress is associated with a non-egress space in the running lifecycle stage"

I did get all the apps in my listings, and I can't help wondering whether that's just because the lists of apps are short!

bin/cloudgov/audit/cf-components
```diff
@@ -0,0 +1,279 @@
...
+      DOMAINCREATED=$(echo "$DOMAIN" | jq -r '.created_at')
+      DOMAINUPDATED=$(echo "$DOMAIN" | jq -r '.updated_at')
+      DOMAININTERNAL=$(echo "$DOMAIN" | jq -r '.internal')
+      DOMAINPROTOCOLS=$(echo "$DOMAIN" | jq -c '.supported_protocols')
+      echo "$level3 $DOMAINNAME (Last updated: $DOMAINUPDATED | Created: $DOMAINCREATED | Internal: $DOMAININTERNAL | Supported Protocols: $DOMAINPROTOCOLS)"
+    done
+  fi
+
+  if [ $spaces == "all" ]; then
+    SPACEPAGES=cf curl "/v3/spaces" | jq -r '.pagination | .total_pages';
```
akf: I'm getting this error message, just once: 
```
curl: (3) URL rejected: No host part in the URL
```
... and I think this is the line it's coming from a line like this where the `cf curl` command isn't enclosed in `$()`. Note that this also applies to line 106.

bin/cloudgov/audit/cf-components
```diff
@@ -0,0 +1,270 @@
...
+    SPACEPAGES=$(cf curl "/v3/spaces" | jq -r '.pagination | .total_pages');
+    if [[ $SPACEPAGES -gt 1 ]]; then
+      for i in $SPACEPAGES; do
+        SPACEIDS=$(cf curl "/v3/spaces?page=$i&per_page=50" | jq -r '.resources[] | select( .relationships.organization.data.guid == "'$ORGID'" ) | .guid')
+      done
+    else
+      SPACEIDS=$(cf curl "/v3/spaces" | jq -r '.resources[] | select( .relationships.organization.data.guid == "'$ORGID'" ) | .guid')
+    fi
+  else
+    SPACEPAGES=cf curl "/v3/spaces" | jq -r '.pagination | .total_pages';
```
akf: Another error here `curl: (3) URL rejected: No host part in the URL` -- I ran the script with a list of spaces specified. 
(It didn't actually affect my results because there weren't enough spaces to require pagination.)

akf (changes requested):
Just one more change -- see my inline comment.

akf (approved):

PR #1803, opened 2024-08-07
USAGOV-1812-wizard-checked-property-vs-class: Rely on the wizard <input> element checked property rather than the checked class

mdranove:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1812

## Description

Update wizard.scss to look for checked property instead of class, and delete wizardChecked.js.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ X ] Frontend (Twig, Sass, JS)
image
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

## Testing Instructions

1. Navigate to any wizard page other than the one with the question "Where did the scam take place?" that has an options list.
2. Check one of the options.
3. Open devtools and inspect the <label> element of the option clicked in step 2.
4. The element does not have the checked class applied. 
5. Close devtools. Look at the element selected in step 2.
6. The element appears with the styling shown in the screenshot above, and the other options appear as expected as well.
7. Hover over the options on the page.
8. The hover color on the unselected options is a grayish color, and the color on the selected option does not change on hover.
9. Refresh the page.
10. Using the keyboard, tab into the options list and change selections by pressing the keyboard's up and down arrows.
11. The styling appears as expected based on the screenshot above.
12. Repeat and validate step 8.
13. Click the "next" button.
14. The next button behaves as expected.
15. On another page with options, click on the "next" button without making a selection.
16. The error message appears as expected.

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

7. Find the commit of this pull request.
17. Build and deploy the changes.
18. Update the Jira ticket by changing the ticket status to `Review in Test` and add a comment. State whether the change is already visible on [cms-dev.usa.gov](http://cms-dev.usa.gov/) and [beta-dev.usa.gov](http://beta-dev.usa.gov/), or if the deployment is still in process.


akf (commented):
I'm having problems with the keyboard navigation, but I am also having the same problem with the current dev branch: 
* I can tab to the first option, but when I tab again I go straight to the "Previous" button. If I then shift-tab, I hit the last option, and then next shift-tab skips past the rest of the options.
* Having tabbed to an option, it appears to be selected, but when I click "Next" I get the error message as if I had not selected an option. 

On that second point, if I hit the space bar after landing on an option, I don't get an error when I click "Next." I'm not the accessibility expert, but I normally would expect a form option not to be checked when I just tab into it.

Again, I'm seeing the same problem on the dev branch, so it's not a problem with this PR.

mdranove:
> I'm having problems with the keyboard navigation, but I am also having the same problem with the current dev branch:
> 
> * I can tab to the first option, but when I tab again I go straight to the "Previous" button. If I then shift-tab, I hit the last option, and then next shift-tab skips past the rest of the options.
> * Having tabbed to an option, it appears to be selected, but when I click "Next" I get the error message as if I had not selected an option.
> 
> On that second point, if I hit the space bar after landing on an option, I don't get an error when I click "Next." I'm not the accessibility expert, but I normally would expect a form option not to be checked when I just tab into it.
> 
> Again, I'm seeing the same problem on the dev branch, so it's not a problem with this PR.

Hi @akf , to the first comment I think that is actually the expected accessibility behavior. Tabbing into the radios should take you to the radio group, but then to navigate through the group you need to use the left/down or right/up arrows on the keyboard. See [W3 radio accessibility guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/radio/).

As to the other comments, yes that's not right. I just pushed up a commit which fixes the issue you're seeing where when you tab into the radios, the first option appears "checked". What was happening was that the the "checked" mixin was getting applied on "focus", which should not have been the case. I've updated the scss so that a special "focus" styling gets applied when you tab in, and the user isn't misled into thinking they've "checked" the radio.

akf:
> Hi @akf , to the first comment I think that is actually the expected accessibility behavior.

You're right and I used to know that, I just keep forgetting!

akf (approved):
Works and looks as described

PR #1839, opened 2024-08-15
USAGOV-1865-published-pages-csv fixes

omerida:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1865

## Description

Wait to get a file lock before writing to CSV file.
Skip raw node paths somehow added by tome after initial export when a redirect is misconfigured

## Type of Changes

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

## Testing Instructions

Generate a tome export and check the CSV file.

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- CSV file should not have any node/* paths which does happen with the DB snapshot USAGOV-1804.prod.10462.post-deploy.sql.gz

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

### DIFF
--- a/web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/PublishedPagesSubscriber.php
+++ b/web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/PublishedPagesSubscriber.php
@@ -32,7 +32,22 @@ class PublishedPagesSubscriber implements EventSubscriberInterface {
     $csv = [];
     $fp = fopen($csv_path, 'c+');
     if ($fp != FALSE) {
-      flock($fp, LOCK_EX);
+
+      $locks = 0;
+      // Wait of 0.01 seconds to get a lock. Try up to 50 times.
+      // Failure to get a lock might is likely extremely rare, but more likely
+      // if using more tome processes to export a file. If we don't get a
+      // lock here, we may overwrite another processes' changes to the CSV.
+      while (FALSE === flock($fp, LOCK_EX)) {
+        $locks++;
+
+        if ($locks > 50) {
+          trigger_error("Could not get lock to update CSV file");
+          break;
+        }
+        usleep(10000);
+      }
+
       while (($line = fgetcsv($fp)) != FALSE) {
         $csv[] = $line;
       }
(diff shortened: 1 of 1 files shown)


web/modules/custom/usagov_ssg_postprocessing/src/EventSubscriber/PublishedPagesSubscriber.php
```diff
@@ -32,7 +32,22 @@ public function modifyHtml(ModifyHtmlEvent $event) {
...
     $fp = fopen($csv_path, 'c+');
     if ($fp != FALSE) {
-      flock($fp, LOCK_EX);
+
+      $locks = 0;
+      // Wait of 0.01 seconds to get a lock. Try up to 50 times.
+      // Failure to get a lock might is likely extremely rare, but more likely
+      // if using more tome processes to export a file. If we don't get a
+      // lock here, we may overwrite another processes' changes to the CSV.
+      while (FALSE === flock($fp, LOCK_EX)) {
```
akf: I have a question! When @jacobaaronyeager and I put this locking in, we struggled with whether we needed to check and retry like this, and we concluded that we didn't. https://www.php.net/manual/en/function.flock.php says "By default, this function will block until the requested lock is acquired [...]" 

And we did some light experimentation and concluded that was true -- if a file was locked, flock() didn't return until the existing lock was released. 

Are you actually finding that flock() returns false rather than blocking? I can certainly believe we missed something.

omerida: I was confused by flocks versus other mutexes and the old comments on that page. Also from other reading, flock is only an exclusive lock within a process. There may still be a potential issue, but in real-world runs, this may happen so quickly that it's not an issue.

I've reverted this code change.

omerida: The other thing that's annoying me is that tome says its exporting 1820ish pages, but the CSV file has about 200 fewer entries than that. It looks like that initial total includes taxonomy and user paths that aren't actually exported.

akf: I thought flock was exclusive across processes too, at least on linux. I should try harder to get something definitive on that. 

Regarding the page count, I'm sure it includes pages that are later excluded by some custom code I put in to exclude whole directories ... which probably includes close to 200 pages of jsonapi paths.

akf (approved):

PR #1896, opened 2024-08-28
USAGOV-1899 Benefits Redirect Modal

cwacht:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1899

## Description

This adds a modal block-type for CMS users to create modals for the website.
It includes javascript to check the url for a modal query parameter and display a modal if one with a matching name exists.

## Type of Changes

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

## Testing Instructions

- Visit the Content blocks page `/admin/content/block`
- Click `+ Add content block`
- Select `Modal`
- Fill in the fields for the modal.
   - `Modal ID` will be used in the URL query parameter so keep it URL safe
   - `Button Text` can be anything since this button remains hidden
   - `Yes Button Text` will be displayed as the button to exit the modal
   - `No Button Text` should be left blank
   - Make sure you select the correct language for the banner.
- Click `Save`
- Visit any page of the site and add "?modal=`modal_id`" to end of the URL
- [ ] Confirm that the modal is displayed properly at any screen size
- [ ] Confirm that the URL parameter is automatically removed from the URL
- [ ] Confirm that you can close the modal
- [ ] Confirm that the modal does not reappear when navigating the site (Even with the browser's back button)
- [ ] Confirm that a spanish modal will only appear on spanish pages and english only on english pages

### Change Requirements

- [x] Requires New Documentation (Link: {})
- [x] Requires New Config
- [x] Requires New Content

### Validation Steps

- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


cwacht:
It would be nice if we could avoid the need to use this div to associate the modal-name with its numerical ID assigned by Drupal's paragraph system.

https://github.com/usagov/usagov-2021/blob/4ef89d0fc6d523237bd17f1e9a755a8080b8cfed/web/themes/custom/usagov/templates/region--header-top.html.twig#L37-L39

As far as I can tell, this would require us to call our modal template directly rather than letting Drupal render the content block entity with Drupal's internally assigned IDs (which currently happens on line 37 above).
I personally have a hard time navigating the data structures to find all the field values to provide them to the template from here:

https://github.com/usagov/usagov-2021/blob/4ef89d0fc6d523237bd17f1e9a755a8080b8cfed/web/modules/custom/usa_twig_vars/usa_twig_vars.module#L337-L340

YaritzaGarcia:
Hello! I'm still reviewing this but since I noticed that it shares a couple of things with the site banner, I wanted to let you know that I changed the way site_banners are added completely because of the cache issues. If this works we should be fine but if you're curious here is the PR: https://github.com/usagov/usagov-2021/pull/1769

config/sync/field.field.block_content.modal.field_modal_id.yml
```diff
@@ -0,0 +1,19 @@
...
+dependencies:
+  config:
+    - block_content.type.modal
+    - field.storage.block_content.field_modal_id
+id: block_content.modal.field_modal_id
+field_name: field_modal_id
+entity_type: block_content
+bundle: modal
+label: 'Modal ID'
+description: 'A name to identify this modal for displaying it programatically. Ex: www.usa.gov/?modal=example_modal_id'
```
akf: Typo: "programmatically" has two 'm's

web/themes/custom/usagov/scripts/modal.js
```diff
@@ -0,0 +1,40 @@
...
+    let modalID = map.getAttribute('data-modal-id');
+    if (modalID === null) return;
+    document.querySelector('[href="#paragraph--id--'+ modalID +'"]').click();
+    document.querySelector('#paragraph--id--'+ modalID).querySelector('button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])').focus();
+}
+
+document.addEventListener('DOMContentLoaded', function() {
+    "use strict";
+
+    document.querySelectorAll('.usa-modal').forEach((modal) => {
```
akf: I'm wondering a couple of things: 

- why set modal.hidden to false in addition to processModalUrlParameter? Or maybe more accurately, why not do this in processModalUrlParameter? It seems to me we'd want to un-hide only the modal referred to in the target. 
- the "toggleModal" function, well, toggles the modal state. This makes me nervous because I've seen to many cases where a toggle function was called when the content was not in the expected state to begin with.

akf (commented):
I put a couple of questions in here, and they are just questions. I haven't had as much time as I'd like to look at this! 

It worked well for me and I wasn't able to reproduce the problem I described with the original site banner implementation, nor am I seeing the problem where Drupal config sync warns that importing the config will delete the blocks I created. We (three) should probably look into why I'm seeing that issue on her PR and not on this one. 

I'm ready to approve this to get it on dev, while we probably do want to follow up and maybe make some additional changes.

web/themes/custom/usagov/scripts/modal.js
cwacht: These are great questions.

### Why set modal.hidden to false:

I was seeing a flash of modal content while the page was loading (similar to the recent gov banner issue), so I made [this commit](https://github.com/usagov/usagov-2021/pull/1896/commits/51b54b2398dd65e6ede9fb613fa6139df665cfc0) to ensure that they start hidden while the page loads. Eventually, the USWDS script hides the modals a different way so we can safely remove our hidden state. Since it is only related to preventing a flash of content, I chose to keep it outside the url parameter function.

### The toggle function:

I agree about the toggle concern, so I've rewritten the script with open and close functions. I also refined everything and added comments for clarity. It still relies on a toggle function because as far as I can tell the USWDS does not provide open and close javascript functions so our options are to trigger their code by clicking their toggle button or try to mimic their code and potentially miss something important.

web/themes/custom/usagov/scripts/modal.js
```diff
@@ -0,0 +1,84 @@
...
+        // Modals start hidden to prevent a flash of content on page load.
+        // This unhides the modals after USWDS script has moved them into a hidden container.
+        modal.hidden = false;
+    });
+
+    let modalName = removeUrlParameter('modal');
```
akf, line 20: Let's sanitize this modalName here. (I think here is better than within "removeUrlParameter." )

web/themes/custom/usagov/scripts/modal.js
akf: Thanks, I understand now. I like your toggle solution.

akf (approved):
I think this is ready for review on dev. There are two things I know of that should be in a follow-on: 
 - Sanitize that modal name after pulling it from the query parameters. I'm thinking that rather than use something like DomPurify, we should just restrict modal names to (letters, numbers, dashes, and underscores). 
 - Put that same restriction in the form for creating/editing the block
 - Default "new revision" to TRUE when saving a block. Actually, don't let people turn that off.

PR #1963, opened 2024-09-19
USAGOV-1728: Site error from a(n intentionally) malformed path

DaleMHFrey:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1728

## Description
Made the usagov redirect module redirect users away from a malformed path that would crash the AliasHandler and to their correct path. This is an edge-case where someone is trying to hit the site directly through index.php in their address-bar. 

## Type of Changes

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions
Before mergin, go to http://localhost/index.phpsssieddgdpathxsx and notice how it crashes the thread. After merging, you must flush the Drupal-cache with bin/drush/cr and then you can try hitting the URL of http://localhost/index.phpsssieddgdpathxsx again - this time the page will load.

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

### DIFF
--- a/web/modules/custom/usagov_redirect/src/PathProcessor/UsagovInboundPathProcessor.php
+++ b/web/modules/custom/usagov_redirect/src/PathProcessor/UsagovInboundPathProcessor.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Drupal\usagov_redirect\PathProcessor;
+
+use Drupal\Core\PathProcessor\InboundPathProcessorInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Processes the inbound path to handle redirects.
+ */
+class UsagovInboundPathProcessor implements InboundPathProcessorInterface {
+
+  /**
+   * Processes the inbound path before resolving the route.
+   *
+   * @param string $path
+   *   The inbound path.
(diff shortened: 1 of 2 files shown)


mdranove (changes requested):
It does what it says, but imo this should be solved with a patch to the autologout module. The stack trace shows exception in autologout.module line 130. Probably just need to wrap it like this:

`  if (isset($array[1])) {
    if (in_array($path_args[1], $paths)) {
      return TRUE;
    }
  } else {
    return;
  }`

This is just my own two cents.

mdranove (approved):
I gave hey this some more thought. I think that this solution is ok since the only path that seems to throw this error is index.php, so solving for that should be enough. Approving.

PR #2098, opened 2024-12-02
USAGOV-2016-v2: Redirect on bad characters in URL

DaleMHFrey:
Redirect when there's a non-breaking space at the end of a request path

https://cm-jira.usa.gov/browse/USAGOV-2016

## Description
Updated the nginx configs to redirect the user away from a URL that ends in a space/thin-space/similar to the same URL without it.
While I was able to test this in the CMS configuration, this is a blind update to the WWW nginx configs, as I am not sure how to test/run the WWW docker container on my local? This test mainly needs to be done on the WWW configs.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions
* Going to a URL of `/foo/bar` should load a not-found page, which is fine
* but after merging this, going to `/foo/bar%C2%A0` should redirect you to `/foo/bar`
* and so should `/foo/bar%E2%80%89` and `/foo/bar%E2%80%83`
* mainly I want to make sure that this configuration is now crashing the WWW nginx container on Dev since this is kind of a blind edit there as I dont quite know how to run the WWW container on my local

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


.docker/src-www/etc/nginx/partials/staticsite.conf.tmpl
```diff
@@ -1,4 +1,21 @@
...
+            set $clean_uri $1;
+            return 301 $clean_uri;
+        }
+        # Match and strip other partial %C2 sequences from the end of the URI
+        if ($uri ~* ^(.+?)(%C2|\xC2)+$) {
+            set $clean_uri $1;
+            return 301 $clean_uri;
+        }
+
+        # try_files $uri @rewrite; # For Drupal <= 6
```
akf: We don't need these try_files lines for the static site. There's no Drupal there and should be no index.php. (Looks like just an accidental copy-n-paste.)

.docker/src-cms/etc/nginx/partials/drupal.conf.tmpl
```diff
@@ -79,6 +79,18 @@
     }
 
     location / {
+        # Decode the URI for proper matching
+        set $decoded_uri $uri;
+        if ($decoded_uri ~* ^(.+?)(%C2%A0|\xA0|%E2%80%89|\xE2\x80\x89|%E2%80%83|\xE2\x80\x83)+$) {
```
akf, line 85: I would like a comment here -- which characters are these?

akf (changes requested):
See comments.

.docker/src-cms/etc/nginx/partials/drupal.conf.tmpl
DaleMHFrey: Ah good idea, this is now done 👍

.docker/src-www/etc/nginx/partials/staticsite.conf.tmpl
DaleMHFrey: oops, you'r right. This is now fixed

akf (approved):

PR #2107, opened 2024-12-05
Usagov 2073 published pages

omerida:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-2073

## Description

Moves generation of the published pages CSV out of the tome export to a drush command that can be executed once the export completes.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [x] Other

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- On dev branch, run tome export and save published-pages.csv as published-pages-good.csv
- Switch to this branch, run tome export. Compare generated published-pages.csv to good branch above. I do this by sorting the files so they have things in the same order (by path and then node id should work), then using a diff tool to compare them.

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


web/modules/custom/usagov_benefit_category_search/src/EventSubscriber/DatalayerAlterSubscriber.php
```diff
@@ -25,7 +25,12 @@ public static function getSubscribedEvents() {
    * Adds category information to the datalayer.
    */
   public function onDatalayerAlter(DatalayerAlterEvent $event): void {
-    $node = \Drupal::routeMatch()->getParameter('node');
+
+    if (!isset($event->datalayer['nodeID']) || !is_numeric($event->datalayer['nodeID'])) {
```
akf, line 31: Is this because the nodeID will not be numeric if it's for a taxonomy page? (If so, I think a comment here would be appropriate.)

web/modules/custom/usagov_ssg_postprocessing/src/Data/PublishedPagesRow.php
```diff
@@ -0,0 +1,216 @@
...
+    // node bundle
+    public readonly string $contentType,
+    // path alias
+    public readonly string $friendlyURL,
+    // node ID
+    public readonly string $pageID,
+    public readonly string $pageTitle,
+    // complete URL including domain name
+    public readonly string $fullURL,
+    public readonly string $taxonomyText1,
```
akf, line 33: I always warn people that the "taxonomy" terms in this report are not Drupal Taxonomy, so I think a comment to that effect would be worthwhile: 

// "Taxonomy" in this report reflects the menu structure of the site (for nodes, this is strictly menu-based)

At this moment, I forget what would be accurate to say about the "wizard" taxonomy terms in this case. Don't sweat it if it's complicated to explain.

scripts/tome-static.sh
```diff
@@ -32,5 +32,13 @@ mkdir -p /var/www/html
 # time drush tome:static -y --uri=$URI --process-count=5 --path-count=1
 time drush tome:static -y --uri=$URI --process-count=$TOME_PROCESS_COUNT --path-count=10
 TOME_SUCCESS=$?
+
 echo "Finished Static Site Generation : "$(date)
+
+if [ "$TOME_SUCCESS" -eq 0 ]; then
+  # path is relative to drupal's web dir
+  time drush usapubcsv modules/custom/usagov_ssg_postprocessing/files/published-pages.csv
```
akf: This call needs `--uri=$URI` (I got "http://default" in my report, which messed with my diffs!)

akf (changes requested):
I found a couple of things. In addition to the couple of items I marked inline, I'm seeing: 
- Differences in the federal_directory_record titles. We compose the title we put in the HTML from the page title plus the acronym in parenthesis, for example "Administration for Children and Families (ACF)". I'm not 100% sure this is a requirement from the Analytics team's point of view (they are the real customer of this report), but it might be. 
- There are also title discrepancies in some of the Wizard pages. For example term 63, where I'm seeing "Su identidad o la de otra persona" in this version of the report and "¿El estafador usó su identidad o se hizo pasar por alguien?" in the dev version. The dev version does reflect what is in the HTML page. It seems that's coming from the Page Title field, which is not a required field, rather than the Name field, which is (and is what you probably expected). 

That Wizard title discrepancy might point to a need to make another field "required" when editing wizard terms, or it might be that we use the Page Title if it exists and otherwise fall back to Name. I didn't look into that. 

I haven't fully reviewed a couple of the files yet, but this is some actionable feedback. 

It is noticeably faster to do the report this way!

omerida:
@akf I've updated to add the comments, include the agency acronym with title, and to use the right title fields for wizard terms. Wizards's use a "heading" field for the on page H1, and fall back to the taxonomy name if it's not set.

akf (dismissed):
Excellent! The only difference this made in the output is that the "Categories" column in the report has a heading, and you've stripped trailing spaces off of some fields, both of which are improvements. 

There's a lint complaint preventing me from merging this: 

```
FILE: .../custom/usagov_ssg_postprocessing/src/Data/PublishedPagesRow.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 180 | ERROR | [x] Expected newline after closing brace
     |       |     (Drupal.ControlStructures.ControlSignature.NewlineAfterCloseBrace)
```

So please update that!

omerida:
sorry about that. I've added a local pre-push git hook to check my code style.

akf (approved):

PR #1670, opened 2024-06-27
USAGOV-1612 Setup a config split to improve developer workflow and disable devel module on production

mdranove:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-1612

## Description
Require and enable config_split module. Configure it to setup a dev split with: 
1. A complete split on the Devel module.
2. A partial split on the autologout.settings to target the timeout and max_timeout config keys.

Disable this split by default, and explicitly enable the split in settings.local.php.

Additionally, enable the admin_toolbar_search module.
## Type of Changes

- [ X ] New Feature
- [ ] Bugfix
- [ ] Frontend (Twig, Sass, JS)
  - Add screenshot showing what it should look like
- [ X ] Drupal Config (requires "drush cim" TWICE)
- [ X ] New Modules (requires rebuild)
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] Egress
  - [ ] Tools
- [ ] Other

## Testing Instructions

**Note that the configuration must be imported twice for the config_split to fully take effect.**

On local environments:

1. Navigate to /admin/config/people/autologout.
2. The Timeout and Max Timeout values are set to 10000
4. Navigate to /admin/config/development/devel.
5. The Variables Dumper setting has the "symfony var dumper" radio option selected

On dev/stg/prod
1. Navigate to /admin/config/people/autologout.
2. The Timeout and Max Timeout values are set to 1800
3. Navigate to /admin/modules. Local the devel module
4. The box is unchecked, indicating that the module is not installed.
6. Find the new admin toolbar search bar.
7. Use it to locate some items in the admin configuration. 
8. The admin toolbar search works as expected, and there are no php errors on the configuration pages you visit.

### Requires New Config
- [ X ] Yes
- [ ] No

### Requires New Content
- [ ] Yes
- [ X ] No

## Security Review

- [ X ] Adds/updates software (including a library or Drupal module)
- [ ] Communication with external service
- [ ] Changes permissions or workflow
- [ ] Requires SSPP updates

## Reviewer Reminders
- Reviewed code changes
- Reviewed functionality
- Security review complete or not required

## Post PR Approval Instructions
Follow these steps as soon as you merge the new changes.

1. Go to the [USAGov Circle CI project](https://app.circleci.com/pipelines/github/usagov/usagov-2021).
4. Find the commit of this pull request.
5. Build and deploy the changes.
[...]

PR #2164, opened 2025-01-09
Usagov 2166 implement config split

whitealbum:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-2166

## Description

implement config_split
devel & devel_generate only available on local development
CSS & JS not aggregated on local development
Clean up settings.php by combining 3 sections with similar logic.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions

Check that aggregation is turned off and Devel (et. al.) is enabled.
Paste 
`
$config['config_split.config_split.local_split']['status'] = FALSE;
$config['config_split.config_split.cloud_split']['status'] = TRUE;
`
after 
`
else {
  $config['config_split.config_split.cloud_split']['status'] = FALSE;
  $config['config_split.config_split.local_split']['status'] = TRUE;
}
`
around line 853.
run `bin/drush cr`. Check [Config Split settings](http://localhost/admin/config/development/configuration/config-split) that 'cloud' is active and 'local' is inactive.
run `bin/drush cim -y`. You should see the aggregation is turned on and Devel should not be enabled.

Then, remove pasted lines, run `bin/drush cr` and `bin/drush cim -y`. The values should have reverted to the original ['local'] state.

### Change Requirements

- [x] Requires New Config
- [ ] Requires New Content

### Validation Steps

- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


akf (commented):
I see this near the top of your description: 

> Currently, only system.performance (aggregation) is working with config_split. I couldn't get devel (et. al.) to work consistently with config_split.

And then under testing instructions you say: 
> Check that aggregation is turned off and Devel (et. al.) is enabled.

It seems that the former is closer to true. What I'm actually seeing is: 

* Devel is checked in /admin/modules
* Devel Generate is also checked (this is not the case with our existing setup)
* The "Devel" button is missing on pages where I'd expect to see it. 

Kinda scratching my head over Devel and Devel Generate being checked, so I filtered by that and hit "Install" on the /admin/modules page, and I got the status message "2 modules have been installed: Devel, Devel Generate" ... hmmm. That restored the "Devel" button but didn't do much for my sense that I know what's going on. 

Anyway, a thing that I see about this changeset is that it removes `devel: 0` from core.extension.yml and doesn't put it back anywhere, and I wonder whether you forgot to include a parallel option to cloud_split. 

The aggregation settings change does work! 

Here's a link to a previous shot at this, in case it sparks any ideas: 

https://github.com/usagov/usagov-2021/pull/1670/files#diff-a2e13b129685d05c8b823bf5e4161647afb41099a55de2c31700338bd81ea300

whitealbum:
@akf This has been improved. See if it is more effective.

web/sites/default/settings.php
```diff
@@ -804,51 +804,54 @@
...
-      $SERVER_HTTP_HOST = 'https://cms-stage.usa.gov';
-      break;
+$IS_CLOUDGOV = FALSE;
+$SERVER_HTTP_HOST = $_SERVER['HTTP_HOST'] ?? 'cms-dev.usa.gov';
+$settings['trusted_host_patterns'] = [];
+$space_name = strtolower($cf_application_data['space_name'] ?? '');
 
-    case "prod":
-      $SERVER_HTTP_HOST = 'https://cms.usa.gov';
+if (in_array($space_name, ['local', 'dev', 'dr', 'stage', 'prod'], true)) {
```
akf: This is giving me ideas. 
 - Can we shorten this up further by treating only `local` and `prod` as special cases, and for all others interpolating the space name into the patterns? The advantage of this would be that if we launch into a new cloud.gov space, we wouldn't have to remember to add a clause for it (as long as we keep to the pattern).  
 - My "local" doesn't have the space_name defined, which probably means most of us don't. We do have the hostnames defined in settings.local.php, which we have in our git repo but explicitly do not include in the images we deploy to cloud.gov. (I don't know what the original reasoning on that was, but I have found it useful to be able to add a local network IP to this file in order to test my local from a mobile device on the same wifi network.) 

That second point is just FYI, although it might argue for eliminating the "local" clause here ... and adding a comment to the effect that local dev relies on settings.local.php for this.

web/sites/default/settings.php
```diff
@@ -804,51 +804,54 @@
...
+  switch (strtolower($cf_application_data['space_name'])) {
+    case "local":
+      $settings['trusted_host_patterns'][] = '^cms-local.usa.gov$';
+      $settings['trusted_host_patterns'][] = '^cms-local-usagov.apps.internal$';
+//      $config['core.extension']['module']['devel'] = 0;
```
akf: Please delete these commented-out lines.

web/sites/default/settings.php
```diff
@@ -804,51 +804,54 @@
...
-$IS_CLOUDGOV=FALSE;
-$SERVER_HTTP_POST = $_SERVER['HTTP_HOST'] ?? 'cms-dev.usa.gov';
-$cf_application_data = json_decode($_ENV['VCAP_APPLICATION'] ?? '{}', TRUE);
-if (!empty($cf_application_data['space_name']) &&
-    in_array($cf_application_data['space_name'],
-             ['dev', 'stage', 'prod'])) {
-  switch (strtolower($cf_application_data['space_name'])) {
     case "dev":
-      $IS_CLOUDGOV=TRUE;
+    case "dr":
```
akf: "dr" needs to have trusted_host_patterns '^cms-dr.usa.gov$' and '^cms-dr-usagov.apps.internal$'

(I was scratching my head over how cms-dr.usa.gov was working at all, but I've figured out that we simply don't have trusted host settings enabled there currently.)

config/sync/config_split.config_split.local_split.yml
```diff
@@ -0,0 +1,18 @@
...
+label: local_split
+description: ''
+weight: 0
+stackable: false
+no_patching: false
+storage: folder
+folder: /var/www/config/local_split
+module:
+  devel: 0
+  devel_generate: 0
```
akf, line 15: We don't currently enable devel_generate. I'm okay with it, especially given that it will be local-only, but wanted to make sure it's intentional.

akf (changes requested):
This time it worked well for me! I made several comments here. The one that needs addressing before merge is that, in the process of cleaning up the code, you've added the trusted_host_patterns setting in the "dr" space. (Omitting this was surely an oversight) Those patterns need to match the space name -- cms-dr, not cms-dev. 

The rest are discretionary.

config/sync/config_split.config_split.local_split.yml
whitealbum: it was intentional.

web/sites/default/settings.php
whitealbum: done

web/sites/default/settings.php
whitealbum: done

web/sites/default/settings.php
whitealbum: let's chat about this.

akf (approved):
Looks good and is working well.

PR #2233, opened 2025-02-20
USGOV-2230 Contact Elected Officials Paragraphs

omerida:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-2230

## Description

Implements the Contact Elected Officials forms and result pages as paragraph entities that don't rely on the "custom twig template" implementation.

## Type of Changes

- [x] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions

**Note** There's a change to [.docker/src-cms/etc/nginx/partials/ssl.conf](https://github.com/usagov/usagov-2021/pull/2233/files#diff-0c1209bd355e436f60726636bcff5ce3d461be82a0bb98a9ee3b0e9afe98135d) needed to make the paragraphs popup display on local. AS far as I can tell, the `X-Frame-Options DENY;` is not what is set on dev/stage/prod. You can rebuild the cms container or ssh into it, make that one line change for testing, save and restart the container.

A config import should bring in the new entities. 

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

Setting up the 3 pages is involved (CEO form, results page, and contact by email form) and documented [here](https://docs.google.com/document/d/1VzTEJ4AHQu65G2CRiAJLtxmLh11F2ptByyszGMal_WA/edit?usp=sharing). To test, follow the instructions to update each page (optionally, you could set up the paragraphs on new pages)

- Confirm CEO form works: errors shown when fields are missing, values from paragraphs form are used, form submits.
- Confirm CEO results works: values from paragraphs form are used in the UI and the contact by email button lead to 3rd page.
- Confirm Contact via Email form works: errors shown when fields are missing, values from paragraphs form are used, form submits.

## Security Review

     * Adds the twig field value module to simply outputting things.
- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


akf:
> AS far as I can tell, the X-Frame-Options DENY; is not what is set on dev/stage/prod.

You've nerd-sniped me here. As so often happens, the presence of this header was a surprise to me, I know some things about what you're seeing, and as I dug in further I ran into a bunch of things that don't seem to make sense. So let's ignore that last bit! 
- This is a good change. 
- The reason a lot of us haven't run into this is that we just use http locally. I think there was a period when it wasn't possible to override the browser warning about a bad certificate; not sure about that. 
- This ssl.conf doesn't come into play in the cloud-deployed app! The WAF app gets the request and proxies it to an http address for the CMS. 

I will actually review your PR soon.

config/sync/user.role.site_admin.yml
```diff
@@ -8,6 +8,7 @@ dependencies:
     - taxonomy.vocabulary.uswds_breakpoints
     - taxonomy.vocabulary.uswds_classes
   module:
+    - action
```
akf: Aha! This change (adding `-action` in user.role.site_admin.yml) needs to go. It's probably the result of a bad merge or something like that. 
Removal of the `action` module (Actions UI) has already been merged to dev, although it was not in the database snapshot I was working with, and this caused `drush cim` to fail with
```
[error]  Drupal\Core\Config\ConfigImporterException: There were errors validating the config synchronization.
Configuration <em class="placeholder">user.role.site_admin</em> depends on the <em class="placeholder">Actions UI</em> module that will not be installed after import. in Drupal\Core\Config\ConfigImporter->validate() (line 823 of /var/www/web/core/lib/Drupal/Core/Config/ConfigImporter.php). 
```

config/sync/system.logging.yml
akf, line 1: Probably an unintentional change from "hide" to "verbose" here?

config/sync/samlauth.authentication.yml
akf, line 1: Please remove the changes to this file. They're different for each server and we don't use this at all for local dev. Nonetheless, the bootstrap.sh script updates this configuration and I guess it defaults to "cms.usa.gov." 

(This is tech debt now. We used to have SecureAuth there as a fallback, but SecureAuth has been decommissioned. So that's another one for the backlog.)

akf (changes requested):
I have more reviewing to do, but I ran into a snag during `drush cim` and found a stray bit of config that probably didn't merge correctly, plus a couple other things that just caught my eye.

akf (commented):
I'm seeing that the *Error alert heading* in **CEO Form, English** is not being used. (I was suspicious because the instructions show the example "Your information contains 4 errors" and it wouldn't always be 4.)

omerida:
> I'm seeing that the _Error alert heading_ in **CEO Form, English** is not being used. (I was suspicious because the instructions show the example "Your information contains 4 errors" and it wouldn't always be 4.)

Ah, these strings are hard coded in ceoSubmission.js, starting at line 360. Since these are dynamic strings, would we want two fields - one for the single error message and another fields that expects a placeholder to be replaced on the fly? 

So, the multiple errors input field would have instrucions like "The placeholder {count} will be replaced with the count of errors." and the expected input is "Your information contains {count} errors" ?

akf:
> Ah, these strings are hard coded in ceoSubmission.js, starting at line 360. Since these are dynamic strings, would we want two fields - one for the single error message and another fields that expects a placeholder to be replaced on the fly?
> 
> So, the multiple errors input field would have instrucions like "The placeholder {count} will be replaced with the count of errors." and the expected input is "Your information contains {count} errors" ?

That sounds good. (I would also be okay with leaving them hard-coded.) 

You've also created fields for some legacy messages, and I think it makes sense to remove those (and simplify the corresponding Paragraph form). None of the fields in the CEO email form are required anymore, so we never display the messages for "topic," "about," and "action." (I see that those are defined both in the javascript in the Footer HTML and in the script ceoSubmission.js, but never used.)

omerida:
> That sounds good. (I would also be okay with leaving them hard-coded.)
> 
> You've also created fields for some legacy messages, and I think it makes sense to remove those (and simplify the corresponding Paragraph form). None of the fields in the CEO email form are required anymore, so we never display the messages for "topic," "about," and "action." (I see that those are defined both in the javascript in the Footer HTML and in the script ceoSubmission.js, but never used.)

I'll leave them hard coded for now, less chance of messing up the JS.

I was wondering if some of those messages were still used, meant to be displayed or if it was a bug that they weren't shown.

akf:
> I was wondering if some of those messages were still used, meant to be displayed or if it was a bug that they weren't shown.

I remember there being a discussion about the required fields on that form. We started off with all the free-form fields being required, and I might have been the one who complained. Ultimately we decided not to require those fields. The form opens a draft in your email client (if your browser is even configured to open mailto links) and if it's blank, that's fine.

web/themes/custom/usagov/scripts/ceoSubmission.js
```diff
@@ -56,7 +56,14 @@ const usps_translations = {
         "invalid-zip": "Por favor, escriba un código postal válido de 5 dígitos."
     }
 };
-let usps_content = usps_translations[document.documentElement.lang];
+
+let usps_content;
+if (usagovUSPSErrors) {
```
akf: I chose to test for backwards-compatibility by attempting to use the CEO tool with the custom twig content. I ran into an error here because usagovUSPSErrors isn't defined. (line 61 of ceoSubmission.js if these comments aren't lining up well with the code) 

So I think we need a check for `(typeof usagovUSPSErrors !== 'undefined')` here. I don't want to make the person deploying this go through converting the form over at deployment time for prod. Especially since that's probably going to be me.

akf (changes requested):
I'm doing backward-compatibility testing on the Spanish forms. I can't submit the form on /functionarios-electos because the form name ("myform") in the custom twig file doesn't match what the script is expecting ("ceo-form"). I see that you did make that update in the English custom twig file.

Sigh, and now I remember that we've resorted to putting raw HTML in the Body of the page again, so it needs to be changed there. We can just add a note to the deployment ticket about that. It still makes sense to update the custom twig file, although I expect we'll remove it in the near future. 

I think that's going to be all of my feedback. I confess that I didn't go through all the form fields in Spanish, so I might do that. But with these last couple of changes, I think this will be ready to go.

omerida:
I fixed the null instance and updated the script to work with both `ceoform` and `myform` IDs. I tested the current English and Spanish forms.

akf (approved):
:tada:

PR #2247, opened 2025-02-26
Usagov 2220 api proxy python

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-2220

## Type of Changes

- [X] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [X] Other

## Testing Instructions
Testing instructions found in the bundled .buildpack/api-proxy/readme.md.

## Security Review

- [x] Communication with external service
- [x] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


bin/cloudgov/deploy-api-proxy
```diff
@@ -0,0 +1,96 @@
...
+    cf create-route "${domain}" --hostname "${hostname}"
+    new_routes+=("${hostname}.${domain}")
+  fi
+}
+
+# Ensure routes exist before deployment
+echo "🔄 Checking routes..."
+ensure_route "apps.internal" "${ORG}-${SPACE_NODOT}-api-proxy"
+ensure_route "app.cloud.gov" "${ORG}-${SPACE_NODOT}-api-proxy"
+ensure_route "usa.gov" "apiproxy"
```
akf: We'll need to make this space-specific. I think what we'll want for dev, stage, and prod are: 
 - apiproxy-dev.usa.gov
 - apiproxy-stage.usa.gov
 - apiproxy.usa.gov
(I'm on the fence about whether "apiproxy" should be "api-proxy".) 

The script fails here if I try to deploy into my sandbox. :-)

bin/cloudgov/deploy-api-proxy
```diff
@@ -0,0 +1,96 @@
...
+{
+  echo "🔑 Configuring Key Storage Service."
+  if service_exists "key-storage" ; then
+    echo "✅ Storage already created."
+  else
+    yes '' | cf create-user-provided-service key-storage -p "{}"
+  fi
+}
+
+# Bind key storage service to the API Proxy
```
akf: This is out of order -- you try to bind the key storage to the api-proxy app before you've verified that the api-proxy app exists ("Deploy API Proxy" step starting on line 67).

bin/cloudgov/deploy-api-proxy
```diff
@@ -0,0 +1,96 @@
...
+  cf restart api-proxy
+  echo "✅ Key Storage Service bound successfully."
+else
+  echo "❌ Failed to bind Key Storage Service to API Proxy."
+  exit 1
+fi
+
+# Deploy API Proxy
+echo "🚀 Deploying API Proxy..."
+if cf push -f .buildpack/manifest-api-proxy.yml --no-route ; then
```
akf: I learned an unpleasant lesson about `--no-route` recently: it will un-map existing routes! 
Yes, I see that the script goes on to map the routes, but basically this causes a momentary outage (which will be more than momentary if the rest of the deployment fails for some reason).

Here's what I did for the logshipper when I discovered this: 

https://github.com/usagov/usagov-logshipper/pull/21/files

The other way we could deal with this would be to update the manifest at build time with the correct routes.

akf (commented):
Going to give you my first batch of feedback and take a break. This is all about the deployment script. 

 - I added some notes at specific lines in the deploy-api-proxy script. 
 - I moved things around in that script so I could continue testing, and got to the point where I deployed the app but it crashed with this error message -- it looks as if you changed the name of the user-provided-service from `credentials` to `key-storage` (or vice versa) at some point and didn't get all the code updated: 
```
   2025-02-26T15:09:54.89-0800 [APP/PROC/WEB/0] ERR File "/home/vcap/app/app/app.py", line 48, in <module>
   2025-02-26T15:09:54.89-0800 [APP/PROC/WEB/0] ERR KEY_STORAGE = VCAP_JSON["user-provided"][0]["credentials"] # Get credentials
   2025-02-26T15:09:54.89-0800 [APP/PROC/WEB/0] ERR KeyError: 'user-provided'
```
 - README.md needs a going-over. I've been following it, somewhat, to test and it still refers to a test-client app, which it either doesn't launch anymore or I'm missing something. 

Also, looking at all the services we deploy,  I also think we should name the key-storage/credentials service in a way that makes it really easy for us know it's for the api-proxy. How about "api-proxy-key-storage"?

.buildpack/api-proxy/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+#### 1) Egress exception
+
+For each api you wish to communicate with, you must add the domain to the egress whitelist by adding it to `bin/cloudgov/apps-egress-allow.acl`.  Be as specific as makes sense for the api; for example, for NASA's api, I added `api.nasa.gov`.  This requires you to run `bin/cloudgov/setup-egress-for-space` on the space that the api-proxy lives to update the acl in the egress.  Then, you either need to re-deploy or run `bin/cloudgov/setup-egress-for-apps` to rotate the proxy passwords for the attached apps.
+
+#### 2) Add key to key store
+
+You **MUST** push a key to the key store on the space the api-proxy lives on with the following data for the API you wish to call:
+
+- `NAME` - API Key Name
```
akf: Please add "Arbitrary name for the key. You will use this in API calls to the api-proxy server."

.buildpack/api-proxy/app/app.py
```diff
@@ -0,0 +1,117 @@
...
+# Load API configuration
+VCAP_SERVICES = os.getenv("VCAP_SERVICES")  # VCAP_SERVICES
+VCAP_JSON = json.loads(VCAP_SERVICES) # Convert to JSON
+if "user-provided" in VCAP_JSON and VCAP_JSON["user-provided"] and "credentials" in VCAP_JSON["user-provided"][0]:
+    KEY_STORAGE = VCAP_JSON["user-provided"][0]["credentials"]  # Get credentials
+else:
+    logger.error("No credentials found in VCAP_SERVICES")
+    KEY_STORAGE = {}
+
+@app.route("/proxy", methods=["GET", "POST", "PUT", "DELETE"])
```
akf: We might need to add "OPTIONS" here.

bin/cloudgov/deploy-api-proxy
```diff
@@ -0,0 +1,134 @@
...
+  echo "❌ Failed to configure network policies."
+  exit 1
+fi
+
+# Bind shared egress group
+echo "🔗 Binding $APP to shared egress..."
+$SCRIPT_DIR/setup-egress-for-apps $APP
+
+PROXYROUTE=$(cf env $app | grep PROXYROUTE | awk '{print $2}')
+if [ -n "$PROXYROUTE" ]; then
```
akf: These need to happen within the app's code, not on the app itself.  The thing that could (and will) go wrong with this approach is that when the app is restaged for any reason, it needs to have these http_proxy variables un-set, and that could happen for reasons other than deployment. 

(I am actually unsure whether "restage" or "restart" is closer to what happens when cloud.gov pushes out an update to their Cloud Foundry platform.) 

You can test this by running "cf restage api-proxy" and watch it fail to install the libraries it needs because of the proxy. 

I think you can do this in a `.profile` file right in the `api-proxy` directory.

akf (changes requested):
I have a few comments here and just a couple of things that I think should be addressed before we merge into dev (the other comments might be good follow-on work). 

Need before merging: 
 - Move the setting of HTTPS_PROXY &c into `.profile` (see my comment for why). I'm pretty sure `.profile` is the right place for it; it's where we do this in other buildpack apps. 
 - Decide whether we're going to rename the "key-storage" service to something like "api-key-storage", and if so, do it. (I'm pushing for this.) 

Follow-on things: 
- What scanning can we add for the python code? Dependabot should take care of dependency scanning, once we merge into dev. Should we be doing anything more? 
- Anything else I commented on that looks worth doing but isn't quick, I guess.

.buildpack/api-proxy/app/app.py
```diff
@@ -0,0 +1,123 @@
...
+        # Remove proxy-specific parameters
+        del params["endpoint"]
+        del params["keyname"]
+
+        logger.info("Forwarding %s request to %s with params %s", method, API_ENDPOINT, params)
+
+        try:
+            response = requests.request(method, API_ENDPOINT, params=params, json=data, headers=headers, timeout=10)
+            logger.info("API response status: %s", response.status_code)
+            sanitized = response.json()
```
akf, line 138: Let's remove `bleach`, it's deprecated: https://github.com/mozilla/bleach?tab=readme-ov-file#bleach

It's also not actually solving a problem here. If you were getting an error flagged on the response, the right thing to do here is probably to find a way to mark that usage as expected -- we wouldn't want to escape the response from the external API like this.

.buildpack/api-proxy/app/app.py
```diff
@@ -0,0 +1,123 @@
...
+                exec(open("extensions/" + filename).read())
+
+        # Remove proxy-specific parameters
+        del params["endpoint"]
+        del params["keyname"]
+
+        logger.info("Forwarding %s request to %s with params %s", method, API_ENDPOINT, params)
+
+        try:
+            response = requests.request(method, API_ENDPOINT, params=params, json=data, headers=headers, timeout=10)
```
akf: I ran the code scans and I'm seeing an "Unsanitized input" flagged here. (It's nice to see that this scan works!)  Looking back at where `API_ENDPOINT` comes from, I think  `params["endpoint"]` would be the thing it's complaining about, since the `API_DOMAIN` is already being taken from a known list. 

I think this is a valid concern! One way to approach this would be to extend the data structure in KEY_STORAGE to include the list of endpoints for each API. (That is, the endpoints we expect to use.) 

That's not a full answer; I think we should discuss it further.

akf (changes requested):
Okay, requesting changes again:

- remove use of `bleach`; it's deprecated (and I don't think what you're bleaching should be bleached)
- I think the scans are complaining about trusting the `endpoint` parameter (the path). Probably a good point; let's think about what we can do here.

.buildpack/api-proxy/requirements.txt
```diff
@@ -0,0 +1,7 @@
+flask
+Flask-Limiter
+gunicorn  # For production server handling
+nh3
```
akf, line 4: It looks like we wound up not using nh3 -- can we remove it?

akf (approved):
Working well enough to merge, thank you!

PR #2267, opened 2025-03-10
USAGOV-2294: Enable us to quickly unpublish all Spanish pages

DaleMHFrey:
Enable us to quickly unpublish all Spanish pages

https://cm-jira.usa.gov/browse/USAGOV-2294

## Description
Create a new module that provides a Drush command for un-publishing all Spanish content.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [x] New Modules
- [ ] Documentation
- [ ] Infrastructure
- [ ] Other

## Testing Instructions
* Merge the code in
* Turn on the new module with `bin/drush en usagov_unpub_es`
* Flush cache after that with `bin/drush cr`
* Run the `unpubspanish` command provided by that new module with `bin/drush unpubspanish`. If the command is not recognized, flush the Drupal ache again.
* Confirm you see a message talking about un-publishing taxonomy-terms and nodes.
* Now delete the `<GitRoot>/html` directory if it exists, and run the static-site-generator with `bin/static-site`
* After that runs, the only `index.html` files found in the `<GitRoot>/html/es/` directory should be redirect pages.

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


web/modules/custom/usagov_unpub_es/src/Drush/Commands/UnpubCommand.php
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  public function unpubSpanishNodes() {
+
+    // Get all Spanish Node-IDs
+    $database = Database::getConnection();
+    $query = $database->select('node_field_data', 'n');
+    $query->fields('n', ['nid']);
+    $query->condition('n.langcode', 'es', '=');
+    $nids = $query->execute()->fetchCol();
```
omerida: We should use entity query here, just like with the terms above. I think the only thing that needs to change is `\Drupal::entityQuery('node')`

web/modules/custom/usagov_unpub_es/src/Drush/Commands/UnpubCommand.php
```diff
@@ -0,0 +1,93 @@
...
+
+    // Calculate total, and verbose info
+    $total = count($nids);
+    print "Going to un-publish {$total} nodes.\n";
+
+    // Un-publish each node
+    $index = 0;
+    foreach ($nids as $nid) {
```
omerida, line 79: I just learned this on another projects, but you can show a progress bar really easily in drush commands. Feedback is appreciated when one of these commands runs. You need this before your loop

```
    $progressBar = new ProgressBar($this->output, $max);
    $progressBar->start();
```

This inside your loop, at the end:
```
$progressBar->advance();
```

And this right after your loop ends.

```
    $progressBar->finish();
```

omerida (dismissed):

web/modules/custom/usagov_unpub_es/src/Drush/Commands/UnpubCommand.php
DaleMHFrey: Cool, this is now implemented.

web/modules/custom/usagov_unpub_es/src/Drush/Commands/UnpubCommand.php
DaleMHFrey: good catch. This is now refactored.

akf (changes requested):
It seems to miss the directory records, both federal and state. These are nodes of type `directory_record` and `state_directory_record`. 

It also didn't unpublish the Spanish translation of the home page for me, even though I do see the code for that here.

DaleMHFrey:
> It seems to miss the directory records, both federal and state. These are nodes of type `directory_record` and `state_directory_record`.
> 
> It also didn't unpublish the Spanish translation of the home page for me, even though I do see the code for that here.

Ah good catch. 
Unfortunately we cannot simply unpublished the Spanish version of a node while keeping the English version published. This is due to our use of the "Content Moderation" module, which overrides Drupal's normal ability to [un]publish individual translations. I believe there may be a module out there to work around this, though I don't know if installing a whole module for one node is overkill? So I have adapted the code to delete the one individual `es` revision on node 1.

The directory nodes will not be unpublished as well

akf (changes requested):
Sorry about this ... let's leave the home page alone. I think it will be better if we have a content editor hand-edit this. I'll put details on the Jira ticket.

akf (approved):

PR #2281, opened 2025-03-17
USAGOV-2315: Dockerfiles to dynamically build architecture

DaleMHFrey:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-2315

## Description
Make the Dockerfiles dynamically build architecture, rather than hard-code for amd64.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [x] CMS
  - [x] minio
- [ ] Other

## Testing Instructions
Switch to this branch, run `docker compose build`

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


.docker/Dockerfile-cms
```diff
@@ -206,13 +204,23 @@ RUN sed -E -i "s/:x:${BUID}:/:x:1919:/g" /etc/passwd \
...
       php83-zlib \
       openssl \
-    && mkdir -p /var/www/web/sites/default/files \
-    # s6 supervisor setup
+      && mkdir -p /var/www/web/sites/default/files
+
+# Determine the system architecture and select the appropriate S6 overlay
+RUN export ARCH=`uname -m` && \
+    if [ "$ARCH" = "aarch64" ]; then \
+        echo "Downloading curl -s https://github.com/just-containers/s6-overlay/releases/download/${S6_VERSION}/s6-overlay-aarch64.tar.gz"; \
```
akf: Remove `curl -s` from the comment, since that's not the command you landed on.

akf (changes requested):
I ran into a problem on my GFE Mac (and now I wonder why this worked for my GFE linux). Basically I got a failure from the `wget` command complaining about the certificate. There's no way you could have seen this, but I can tell you what to do. 

Basically, the few lines that start with "# Add the zscaler certificate to the trusted certs only on local builds" starting at line 263? Those need to come before you run wget. And they need to run *after* the apk add command that installs the ca-certificates package. 

Incidentally, I decided to do the full install when I tested: 
```
docker system prune --all
/bin/init
```

Background: GSA-issued computers have ZScaler installed. ZScaler proxies HTTP requests and winds up applying its own certificate (there are exceptions), which is why we install the ZScaler certificate in these images when we build locally. It seems I should have had this problem on my Linux workstation as well but I didn't.

.docker/Dockerfile-cms
DaleMHFrey: Ah good catch. These issues are now resolved on this branch

akf (approved):
I discovered (by running this on my Mac GFE) that the BUILD_ENV variable needed to be added to a stage of Dockerfile-cms that hasn't needed it before now. It seemed petty to hand that back, especially since you don't have a way to test this, so I made the commit. 

It built successfully on the Mac GFE after that.

PR #2285, opened 2025-03-24
USAGOV-2249: Upgrade nginx in CMS And WWW apps

DaleMHFrey:
Upgrade nginx in CMS And WWW apps

https://cm-jira.usa.gov/browse/USAGOV-2249

## Description
Compiling nginx from source to get this version on Alpine.

## Type of Changes
- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [x] CMS
  - [ ] WAF
  - [x] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions
* I suggest erasing all your Docker cache before rebuilding this. In docker Dsahboard, click on the Bug-icon in the top-right area, and then click the "Clean / Purge Data" option
* Then rebuild with `./bin/init` and then `docker compose up`
* You may need to restore the database (if you Purged all the docker data)
* At this point, we just check to confirm the site is still running and that no errors are seen in the `docker compose up` console.

### DIFF
--- a/.docker/Dockerfile-cms
+++ b/.docker/Dockerfile-cms
@@ -205,6 +205,7 @@ RUN sed -E -i "s/:x:${BUID}:/:x:1919:/g" /etc/passwd \
       php83-xmlwriter \
       php83-zip \
       php83-zlib \
+      openssl \
     && mkdir -p /var/www/web/sites/default/files \
     # s6 supervisor setup
     && tar xzf /tmp/s6overlay.tar.gz -C / \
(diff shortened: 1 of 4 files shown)


.docker/Dockerfile-cms
```diff
@@ -275,6 +275,37 @@ RUN chmod -R +x /etc/periodic \
...
+RUN if pgrep nginx; then pkill nginx; fi
+
+RUN apk add --no-cache gcc libc-dev make pcre-dev zlib-dev openssl-dev curl gnupg lua5.1 lua5.1-dev luajit luajit-dev libpcrecpp cmake
+RUN apk del nginx nginx-mod-http-lua || true
+RUN cd /tmp/ \
+    && git clone --branch v0.10.26 https://github.com/openresty/lua-nginx-module.git \
+    && wget https://github.com/simpl/ngx_devel_kit/archive/refs/tags/v0.3.1.tar.gz \
+    && tar -xzvf v0.3.1.tar.gz \
+    && mv ngx_devel_kit-0.3.1 ngx_devel_kit \
+    && wget http://nginx.org/download/nginx-1.27.0.tar.gz \
```
akf: Let's use https, please, and use nginx-1.27.4 (which is the latest 1.27 right now). 
(I think we should do a weekly check for new point versions and just go ahead and make that update on dev.)

And can we check the tarball against its checksum? There might be something worth cribbing from here: https://github.com/nginx/docker-nginx/blob/cffeb933620093bc0c08c0b28c3d5cbaec79d729/mainline/alpine/Dockerfile

(

akf (changes requested):

.docker/Dockerfile-cms
DaleMHFrey: Ah, good Idea! Ok, this is now done

.docker/Dockerfile-www
```diff
@@ -93,6 +93,37 @@ RUN chmod -R +x /etc/periodic \
...
+RUN if pgrep nginx; then pkill nginx; fi
+
+RUN apk add --no-cache gcc libc-dev make pcre-dev zlib-dev openssl-dev curl gnupg lua5.1 lua5.1-dev luajit luajit-dev libpcrecpp cmake
+RUN apk del nginx nginx-mod-http-lua || true
+RUN cd /tmp/ \
+    && git clone --branch v0.10.26 https://github.com/openresty/lua-nginx-module.git \
+    && wget https://github.com/simpl/ngx_devel_kit/archive/refs/tags/v0.3.1.tar.gz \
+    && tar -xzvf v0.3.1.tar.gz \
+    && mv ngx_devel_kit-0.3.1 ngx_devel_kit \
+    && wget http://nginx.org/download/nginx-1.27.0.tar.gz \
```
akf: We need the same changes here as in the Dockerfile-cms file!

akf (changes requested):
**Good:** The cms container built for me and is running nginx 1.27.4. 
**Oops:** We need the same updates to the build file for www that you did for cms. 

And a question: why this particular version of the lua-nginx-module? I'm just curious.
```git clone --branch v0.10.26 https://github.com/openresty/lua-nginx-module.git```

.docker/Dockerfile-www
DaleMHFrey: Ah, can't believe I missed that! Thanks. This is now fixed.

akf (approved):

PR #2297, opened 2025-04-01
Usagov 2314 cron for scheduled publish

jacobaaronyeager:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-2314

## Description
Set up cron app to fire lightweight cron for content scheduler.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [X] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [X] Cron
- [ ] Other

## Testing Instructions
Create a node in draft and set a scheduled time in the future.  Cron runs every 5 mins, so check back on the 5 minute mark after the scheduled time.

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


.docker/src-cron/etc/periodic/5min/run-lightweight-cron.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ "${CF_INSTANCE_INDEX:-''}" == "0" ]; then
+
+  # Use unique uri per environment - default to prod
+  if [ "${APP_SPACE}" = "dev" ]; then
+    URI="http://cms-dev-usagov.apps.internal"
+  elif [ "${APP_SPACE}" = "dr" ]; then
+    URI="http://cms-dr-usagov.apps.internal"
+  elif [ "${APP_SPACE}" = "stage" ]; then
+    URI="http://cms-stage-usagov.apps.internal"
+  elif [ "${APP_SPACE}" = "local" ]; then
```
akf: Does "http://localhost" really work here? (I guess I'll have to try and see.) I would expect it not to work, from the POV of the cron app, and that it might need to be "http://cms-usagov.docker.local", which is what the cypress tests use. 

Also I think we could simplify this by setting the URI to "http://cms-${APP_SPACE}-usagov.apps.internal" for all but local dev.

.docker/src-cron/etc/periodic/5min/run-lightweight-cron.sh
```diff
@@ -0,0 +1,28 @@
...
+    URI="http://localhost"
+  else
+    URI="http://cms-prod-usagov.apps.internal"
+  fi
+
+  echo "Running lightweight cron on $URI/scheduler/cron/$CRON_KEY"
```
akf: Let's not log the $CRON_KEY, even though it's not super-sensitive. It just seems like bad form. Probably we don't this log statement at all now that you've got it working?

docker-compose.yml
```diff
@@ -169,6 +169,39 @@ services:
     networks:
       - gsa-gov
 
+  cron:
+    container_name: cron
+    image: gsatts/usagov-2021:cron
+    mem_limit: 128m
+    storage_opt:
+      size: 4G
```
akf: Do we need to specify storage_opt? Isn't this kind of big? 
Actually, let's take this out. On my Mac, I get an error "--storage-opt is supported only for overlay over xfs with 'pquota' mount option". I think I had to change my filesystem setting because of another bug.

docker-compose.yml
```diff
@@ -169,6 +169,39 @@ services:
...
+    command: ["sh", "-c", "while :; do echo 'Cron running'; sleep 60; done"]
+    networks:
+      - gsa-gov
+    depends_on:
```
akf: Doesn't really depend on minio, does it?

docker-compose.yml
```diff
@@ -169,6 +169,39 @@ services:
...
+      - minio
+    environment:
+      HEALTHCHECK_TYPE: process
+      NO_ROUTE: true
+      VCAP_SERVICES: >
+        { "s3": [{
```
akf: I don't think the cron app should be messing with the "storage" bucket (which is for static files)! 

I'm guessing this is just a holdover from the cms.

akf (commented):
I like the idea of being able to run the cron in Docker, but I'm getting a lot of errors and I wonder whether you abandoned that and forgot to take out the `docker-compose.yml` changes. 

The errors are with other tasks than schedule publish (I think I might not have left it running long enough for that). Here's a sample: 

```
cron  | uploading waittime.json to S3.
cron  | 
cron  | Invalid endpoint: https://s3..amazonaws.com
cron  | Clean task: ccwaittimes.sh / 247
cron  | Created lock /tmp/tasks/run/cfevents.sh/394
cron  | : CF space '2025-02-05t01:10:09z' does not exist!
cron  | 
cron  | Invalid endpoint: https://s3..amazonaws.com
cron  | 
cron  | Invalid endpoint: https://s3..amazonaws.com
```

The Dockerfile and docker-compose will need a bit more work if we're going to keep the local setup.

docker-compose.yml
jacobaaronyeager: I removed the cron section altogether.  I never did get it working and forgot about it.  An oversight on my part.

docker-compose.yml
jacobaaronyeager: Same as above.

docker-compose.yml
jacobaaronyeager: Same as above.

.docker/src-cron/etc/periodic/5min/run-lightweight-cron.sh
jacobaaronyeager: Removed.

.docker/src-cron/etc/periodic/5min/run-lightweight-cron.sh
jacobaaronyeager: This has been simplified and local option has been changed to suggestion.

.docker/src-cron/etc/periodic/5min/run-lightweight-cron.sh
```diff
@@ -0,0 +1,21 @@
...
+
+  # Use unique uri per environment - default to prod
+  if [ "${APP_SPACE}" = "local" ]; then
+    URI="http://cms-usagov.docker.local"
+  else
+    URI="http://cms-${APP_SPACE}-usagov.apps.internal"
+  fi
+
+  echo "Running lightweight cron"
+  curl "$URI/scheduler/cron/$CRON_KEY"
```
akf: Spews too much to the log files (especially when the CRON_KEY isn't set up on the other side.

akf (changes requested):
That curl command needs not to send so much to the logs. I tried adding the `-s` flag (to a curl command I ran in my terminal) and that doesn't silence the output either. We might need to use `-o` and send the output to a file in /tmp. 

This is because we get our whole 403 page when the cron key isn't set up correctly. When it is set up, we still get several lines that aren't useful: 

```
   2025-04-10T12:40:00.58-0700 [APP/PROC/WEB/0] OUT Running lightweight cron
   2025-04-10T12:40:00.58-0700 [APP/PROC/WEB/0] ERR % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
   2025-04-10T12:40:00.58-0700 [APP/PROC/WEB/0] ERR Dload  Upload   Total   Spent    Left  Speed
   2025-04-10T12:40:01.24-0700 [APP/PROC/WEB/0] ERR 0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
   2025-04-10T12:40:01.24-0700 [APP/PROC/WEB/0] ERR 0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
   2025-04-10T12:40:01.24-0700 [APP/PROC/WEB/0] ERR 0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

```

Probably `-s` takes care of that. So I'm thinking `curl -s -o /tmp/scheduler_output "$URI/scheduler/cron/$CRON_KEY"`

akf (approved):
We should delete the feature branch references from .circleci/config.yml, but that can happen later.

PR #2360, opened 2025-05-28
Usagov 1685 fix site banner

jacobaaronyeager:
## Jira Task

## Description
Adds Config Ignore module and ignores block.block.site_banner_* and block.block.usagov_* to avoid overriding blocks generated by the site banner module.

## Type of Changes

- [X] New Feature
- [ ] Bugfix
- [X] Drupal Config (requires "drush cim")
- [X] New Modules (requires rebuild)
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions
Create site banner blocks and then attempt config import.  If the site banner blocks persist, we're golden.

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


akf (changes requested):
I have two issues: 

First, I think the `block.block.usagov_*` pattern is going to cause us problems because we have a lot of blocks that match that pattern and that we do manage as configuration. (Do we need that? I took it out of the "ignore" group and it looks like only `block.block.site_banner_*` have been added.) 

Second, this is what happens for me when I add an English banner according to the instructions: 

- It appears twice on the English pages. (Once would be good)
- It appears once on the Spanish pages. (Zero would be good)

A similar thing happens when I add a Spanish banner. So now I have three banners on every page; two in the language of the current page and one not.

jacobaaronyeager:
The duplicated banner seems to happen when the following checkbox is checked:
Screenshot 2025-05-30 at 11 57 19 AM
It has some odd interaction with the paragraphs module (perhaps this box is generated by paragraphs?) that duplicates it.

akf:
> The duplicated banner seems to happen when the following checkbox is checked: 

That checkbox is defined in [field.field.block_content.site_banner.field_place_above_header.yml](https://github.com/usagov/usagov-2021/blob/dev/config/sync/field.field.block_content.site_banner.field_place_above_header.yml) -- which is part of what @YaritzaGarcia implemented last year to make it possible to add the banners. 

docs/SiteBanner.md explicitly says to check that box to display the banner. (We expect that there will be cases where people will want to disable a banner by un-checking that box so they can easily re-enable it later.)

config/sync/config_ignore.settings.yml
```diff
@@ -0,0 +1,6 @@
+_core:
+  default_config_hash: IgOVnECx6lbVt6JVFnadoEEugneDf3UblPZnOzov43Q
+mode: simple
+ignored_config_entities:
+  - 'block.block.site_banner_*'
+  - 'block.block.usagov_*'
```
akf, line 6: I think the block.block.usagov_* pattern is going to cause us problems because we have a lot of blocks that match that pattern and that we do manage as configuration. (Do we need that? I took it out of the "ignore" group and it looks like only block.block.site_banner_* have been added.)

akf (changes requested):
I'm still getting multiple banners (two in the language I specified and one in the other) for each banner I add. And I still think that `block.block.usagov_*` pattern is unnecessary and will probably give us trouble in the future.

akf:
Closing in favor of #2469

PR #2361, opened 2025-05-29
USAGOV-2390: Make search affiliate strings configurable via CMS

DaleMHFrey:
Make search affiliate strings configurable via CMS

https://cm-jira.usa.gov/browse/USAGOV-2390

## Description
Made the affiliate pull from Drupal-states, so we can configure this via drush.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure\
- [ ] Other

## Testing Instructions
* Switch to this branch and make sure to `bin/drush cr`
* Load the front end on your local, expect nothing to change when you inspect the page and and search for the word "siteHandle"
* Now run `bin/drush state:set usagov_affiliate "your_value_here"`
* do another `bin/drush cr`
* After reloading the front end, expect to find "your_value_here" in the HTML source-code of the page

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


akf (changes requested):
I see that this replaces `usagov_en_all_gov` but not `usagov_es_internal`. Haven't tested it yet, but we want to be able to update both (and they are separate values).

DaleMHFrey:
> I see that this replaces `usagov_en_all_gov` but not `usagov_es_internal`. Haven't tested it yet, but we want to be able to update both (and they are separate values).

oops! Good catch - this is now resolved

web/modules/custom/usagov_menus/src/Plugin/Block/MobileMenuBlock.php
```diff
@@ -42,13 +43,14 @@ public function build(): array {
       case 'en':
       default:
         $menuID = 'left-menu-english';
+        $affiliate = \Drupal::state()->get('usagov_affiliate', 'usagov_all_gov');
```
akf: The default here should be `usagov_en_all_gov`, not `usagov_all_gov`.

web/themes/custom/usagov/html-blocks/header.html
```diff
@@ -32,7 +32,7 @@
 
         <form class="usa-search usa-search--small" action="https://search.usa.gov/search" method="get" name="search_form" accept-charset="UTF-8" role="search">
             <label id="top-srch" class="usa-sr-only" for="search-field-small">Search</label>
-            <input id="affiliate" name="affiliate" type="hidden" value="usagov_en_all_gov">
+            <input id="affiliate" name="affiliate" type="hidden" value="{{ usagov_affiliate_lang }}">
```
akf, line 35: This is coming out with a blank where the value should be for me: 
```
  <input id="affiliate" name="affiliate" type="hidden" value="">
```
I have not yet run drush to set state; I'm just seeing how the defaults pan out. (The correct affiliate code appears next to `siteHandle` and the Spanish affiliate codes seem to be coming through correctly.)

web/themes/custom/usagov/usagov.theme
```diff
@@ -557,3 +561,24 @@ function usagov_preprocess_breadcrumb(&$variables): void {
     $variables['breadcrumb'][] = ['text' => $currentTitle];
   }
 }
+
+function usagov_preprocess_header(array &$variables) {
+  $affiliate = \Drupal::state()->get('usagov_affiliate', 'usagov_en_all_gov');
+  $affiliate_es = \Drupal::state()->get('usagov_affiliate_es', 'usagov_es_internal');
+
+  $current_path = \Drupal::service('path.current')->getPath();
+  $variables['usagov_affiliate_lang'] = str_starts_with($current_path, '/es') ? $affiliate_es : $affiliate;
```
akf: Do we not have the language setting of the page available here? I think that if this were working, it would put the Spanish search on `/estimate-college-cost`.

akf (changes requested):
See my comments. This is working for the bit of code that puts out `siteHandle` but it's not populating the forms for the search boxes correctly. I don't think the preprocess hooks are running. 
(Possibly I've forgotten how to refresh a theme so those run? I've cleared a lot of caches.)

web/themes/custom/usagov/usagov.theme
DaleMHFrey: ah, I have found a better way and have applied this

web/modules/custom/usagov_menus/src/Plugin/Block/MobileMenuBlock.php
DaleMHFrey: oops, fixed

web/themes/custom/usagov/html-blocks/header.html
DaleMHFrey: oh! I have now realized I was using the wrong preproccessor, this is now fixed

DaleMHFrey:
Ah, I was able to reproduce issues mentioned, and they are now resolved. This is ready for another review

akf (changes requested):
Looking much better! This works as desired for me on English pages. 

On Spanish pages, one field is not getting updated. I ran: 
`drush state:set usagov_affiliate_es "hola_que_tal"`

This one input, in the desktop search form, remains stubbornly unchanged: 
```
  <input id="affiliate" name="affiliate" type="hidden" value="usagov_es_internal">
```

I see it! There is a `top_nav_es.html.twig` as well as `top_nav.html.twig`. You've edited the English version but not the Spanish.

DaleMHFrey:
> Looking much better! This works as desired for me on English pages.
> 
> On Spanish pages, one field is not getting updated. I ran: `drush state:set usagov_affiliate_es "hola_que_tal"`
> 
> This one input, in the desktop search form, remains stubbornly unchanged:
> 
> ```
>   <input id="affiliate" name="affiliate" type="hidden" value="usagov_es_internal">
> ```
> 
> I see it! There is a `top_nav_es.html.twig` as well as `top_nav.html.twig`. You've edited the English version but not the Spanish.

Ah, good catch - this is now resolved

akf (dismissed):
Works well, thank you!

akf (approved):
I removed a line of whitespace to make the linter happy. Ideally someone else would review this now because of the additional commit, but we can see it's just a whitespace change.

PR #2484, opened 2025-10-08
Usagov 2535 automate backups

jacobaaronyeager:
## Jira Task

https://gsa-standard.atlassian-us-gov-mod.net/browse/USAGOV-2535

## Description
Automates static site (every run) and public files (if there are changes) backups on every Tome run.
Automate nightly DB backups, scheduled for 7pm EST.

## Type of Changes

- [X] New Feature
- [ ] Bugfix
- [X] Documentation
- [X] Infrastructure
  - [X] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions
Deploy to dev environment.

For static site and public files:
Let tome run. On the box, run `scripts/snapshot/test.sh`. If all tests pass, run `scripts/snapshot/manager.sh list` and make sure there are backups available. You may make a test backup with `./scripts/snapshot/manager.sh backup static,public TEST`.

For DB:
1. Wait until 7pm est for an auto backup to happen OR
2. Run `./scripts/snapshot/manager.sh backup db` to force a backup.  This doesn't test the cron functionality, but it does make sure the backup itself works.

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

3. Find the commit of this pull request.
4. Build and deploy the changes.
5. Update the Jira ticket by changing the ticket status to `Review in Test` and add a comment. State whether the change is already visible on [cms-dev.usa.gov](http://cms-dev.usa.gov/) and [beta-dev.usa.gov](http://beta-dev.usa.gov/), or if the deployment is still in process.


scripts/db-backup-daily.sh
```diff
@@ -0,0 +1,275 @@
+#!/bin/sh
+
+# Automated Database Backup Script
+# Check if database backups are enabled
+if [ "$ENABLE_DB_BACKUPS" != "true" ]; then
```
arpage: @jacobaaronyeager - this is called before the config is read - so seems like it would always fail.

scripts/db-backup-daily.sh
```diff
@@ -0,0 +1,275 @@
...
+    log_message "Database backups disabled"
```
arpage: @jacobaaronyeager - log_message() is not defined yet

arpage (changes requested):
I've reviewed backup functionality, and the site/public stuff seems to work, but the daily db backup script has a couple issues (see review comments in the file).

I think if the changes got in Tuesday before the cron runtime, then we would see if the cron-based db backup works.

I still need to read the restore docs.  This is really cool stuff btw!

arpage:
Setup during bootstrap.sh fails:

```
Setting up automatic backup cron system
/var/www/scripts/bootstrap.sh: line 246: /var/www/scripts/setup-automatic-backup-cron.sh: not found
```

Investigating.

.docker/Dockerfile-cms
```diff
@@ -288,7 +289,13 @@ RUN chmod -R +x /etc/periodic \
     && echo >> /etc/motd \
     && cp /etc/motd /var/www/web/version.txt \
     && echo "PATH=$PATH" > /etc/profile.d/drupal.sh \
-    && echo "alias tll='tail -f \$(ls -tr *.log | tail -1)'" > /etc/profile.d/usagov-aliases.sh
+    && echo "alias tll='tail -f \$(ls -tr *.log | tail -1)'" > /etc/profile.d/usagov-aliases.sh \
+    && echo '#!/bin/sh' > /etc/profile.d/setup-cron.sh \
+    && echo '# Auto-setup database backup cron job on container start' >> /etc/profile.d/setup-cron.sh \
+    && echo 'if [ -f /var/www/scripts/snapshot/setup-cron.sh ]; then' >> /etc/profile.d/setup-cron.sh \
+    && echo '    /var/www/scripts/snapshot/setup-cron.sh setup >/dev/null 2>&1' >> /etc/profile.d/setup-cron.sh \
```
arpage: Does this conflict with (or make unnecessary) the setup-cron.sh call in bootstrap.sh?

arpage:
@jacobaaronyeager - 

I as able to do a `./manager.sh clean db` and successfully remove all db backups.
(edit: same for public - although it deletes the corresponding static snapshot).

However, manager.sh fails when trying to delete snaps based on date (local and container).

Container:
```
3d8bdb39-e5c4-40f7-522d-a616:/var/www/scripts/snapshot# ./manager.sh clean db,static,public 30
⚠️  This will delete db,static,public backups older than 30 days
Continue? [y/N]: y
🧹 Cleaning up backups...
❌ Error: Date calculation not supported on this system
⚠️  Use 'list' command for manual cleanup with AWS CLI.
```

Local:
```
❯ scripts/snapshot/local-manager.sh clean db,static,public 30
⚠️ This will delete db,static,public backups older than 30 days
Continue? [y/N]: y
🧹 Cleaning up backups...
❌ Error: Date calculation not supported on this system
⚠️ Use 'list' command for manual cleanup with AWS CLI.
```

arpage:
@jacobaaronyeager - Can the db backup be removed from the set performed during tome runs?   We wait for tome to stop before doing the db backup, but we're running from tome so it will always wait 25 minutes then die :(

jacobaaronyeager:
> @jacobaaronyeager - Can the db backup be removed from the set performed during tome runs? We wait for tome to stop before doing the db backup, but we're running from tome so it will always wait 25 minutes then die :(

Yep, that was not intentional and a mistake in how I called the backup command.  It's fixed.

akf:
Hi @jacobaaronyeager, I am reviewing our SSPP and we say: 

> The USA.gov team stores production backups in cloud.gov-brokered S3 for 180 days.

I would take that to mean all three kinds of backups (database, static files, public files). The defaults for `*_AUTO_CLEANUP`  here are considerably less than that. Could you have your script check that value when in prod and ensure it's not less than 180 days?

arpage:
@jacobaaronyeager small issue with manager.sh:  if it is run from an unexpected directory location, it announces a failure but it will not exit immediately, and proceeds to list all files in the current directory (e.g. if it is run from a tome-log folder, it will list all the log files there).

arpage:
@jacobaaronyeager - I made a change to the database (added some text to /mark-test), and generated the static site.  New text showed up.  I made a backup of the current state of dev w/ local-manager.

I then used local-manager to restore the 1st (0th) backup of 12-03. The static and public files were restored, but something is off with the images - they are all showing broken in dev.

Also, the database did not restore - 

```
✅ Public files restored
🔄 Restoring database...
⏳ Waiting for Tome to stop (max: 25 minutes)...
✅ Tome has stopped
🔒 Disabling Tome...
✅ Tome disabled: 1
🚧 Enabling maintenance mode...
mktemp: : Invalid argument
✅ Maintenance mode enabled: 1
download: s3://cg-33ba2c88-f377-4249-8b26-0a9d24caf3f5/auto-backups/database/AUTO-dev-14857-2025-12-03-0.sql.gz to ./AUTO-dev-14857-2025-12-03-0.sql.gz
❌ ERROR: Failed to decompress database backup
🚧 Disabling maintenance mode...
✅ Maintenance mode disabled: 0
🔓 Re-enabling Tome...
✅ Tome re-enabled (disabled flag: none
```

I left dev as-is for you to take a look at.

arpage:
The following shows that tome is indeed being disabled while the database backups are run:

```
-rw-r--r--    1 root     root           192 Dec  5 15:55 dev-2025_12_05_15_55_00.log
-rw-r--r--    1 root     root            31 Dec  5 15:56 dev-2025_12_05_15_56_00.log
-rw-r--r--    1 root     root           192 Dec  5 15:57 dev-2025_12_05_15_57_00.log
```

(I had set the cron-based db backup to run at 15:55 - the "192 bytes" sized log is "no changes". The "31 bytes" sized log is "tome is disabled")

arpage:
Hi @jacobaaronyeager 

I was trying to restore only public and static, per the syntax in the help text.  However, it looks like restore wants to restore everything, regardless of the arguments.  This might be just the script output not reflecting what is actually going to happen - not sure:

```
❯ scripts/snapshot/local-manager.sh restore AUTO-dev-14866-2025-12-04-2 --only=static,public
🔄 Checking restore options

✅ Static site backup found: AUTO-dev-14866-2025-12-04-2
✅ Public backup found: AUTO-dev-14866-2025-12-04-2
⚠️ Using closest database backup: AUTO-dev-14857-2025-12-03-0.sql.gz

Restore plan:
Static site:   AUTO-dev-14866-2025-12-04-2
Public files:  AUTO-dev-14866-2025-12-04-2
Database:      AUTO-dev-14857-2025-12-03-0.sql.gz

This will overwrite current data!
Continue with restore? (y/N): n
❌ Cancelled.
❯ scripts/snapshot/local-manager.sh restore AUTO-dev-14866-2025-12-04-2 --only=static
🔄 Checking restore options

✅ Static site backup found: AUTO-dev-14866-2025-12-04-2
✅ Public backup found: AUTO-dev-14866-2025-12-04-2
⚠️  Using closest database backup: AUTO-dev-14857-2025-12-03-0.sql.gz

Restore plan:
Static site:   AUTO-dev-14866-2025-12-04-2
Public files:  AUTO-dev-14866-2025-12-04-2
Database:      AUTO-dev-14857-2025-12-03-0.sql.gz

This will overwrite current data!
Continue with restore? (y/N): n
❌ Cancelled.
❯ scripts/snapshot/local-manager.sh restore AUTO-dev-14866-2025-12-04-2 --only=db
🔄 Checking restore options

✅ Static site backup found: AUTO-dev-14866-2025-12-04-2
✅ Public backup found: AUTO-dev-14866-2025-12-04-2
⚠️  Using closest database backup: AUTO-dev-14857-2025-12-03-0.sql.gz

Restore plan:
Static site:   AUTO-dev-14866-2025-12-04-2
Public files:  AUTO-dev-14866-2025-12-04-2
Database:      AUTO-dev-14857-2025-12-03-0.sql.gz

This will overwrite current data!
Continue with restore? (y/N): n
```

arpage:
@jacobaaronyeager - 

`--only=db` and `--only=public` does not find/restore chosen tag

```
AUTO-dev-14876-2025-12-05-1      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-05-1
AUTO-dev-14876-2025-12-06-0      ✅      ✅      ✅      restore AUTO-dev-14876-2025-12-06-0
AUTO-dev-14876-2025-12-06-1      ✅      ✅      ✅      restore AUTO-dev-14876-2025-12-06-1
AUTO-dev-14876-2025-12-06-10     ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-10
AUTO-dev-14876-2025-12-06-2      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-2
AUTO-dev-14876-2025-12-06-3      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-3
AUTO-dev-14876-2025-12-06-4      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-4
AUTO-dev-14876-2025-12-06-5      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-5
AUTO-dev-14876-2025-12-06-6      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-6
AUTO-dev-14876-2025-12-06-7      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-7
AUTO-dev-14876-2025-12-06-8      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-8
AUTO-dev-14876-2025-12-06-9      ✅      ✅      ❌      restore AUTO-dev-14876-2025-12-06-9
USAGOV-2563-dev-14857-2025-12-02--post-deploy-0 ✅      ✅      ✅      restore USAGOV-2563-dev-14857-2025-12-02--post-deploy-0

✅ = Available    ❌ = Missing (smart fallback may apply)
❯ scripts/snapshot/local-manager.sh restore AUTO-dev-14876-2025-12-06-1 --only=db
🔄 Checking restore options


Restore plan:

This will overwrite current data!
Continue with restore? (y/N): y

🔄 Restoring...

🎉 Restore complete!
❯ scripts/snapshot/local-manager.sh restore AUTO-dev-14876-2025-12-06-0 --only=db
🔄 Checking restore options


Restore plan:

This will overwrite current data!
Continue with restore? (y/N): y

🔄 Restoring...

🎉 Restore complete!
```


```
❯ scripts/snapshot/local-manager.sh restore AUTO-dev-14876-2025-12-06-1 --only=public
🔄 Checking restore options

✅ Public backup found: AUTO-dev-14876-2025-12-06-1

Restore plan:
Public files:  AUTO-dev-14876-2025-12-06-1

This will overwrite current data!
Continue with restore? (y/N): y

🔄 Restoring...
🔄 Restoring public files...
✅ Public files restored
🔄 Refreshing file metadata cache...
✅ File metadata cache refreshed

🎉 Restore complete!
Restored: public files
```

arpage:
@jacobaaronyeager - for some reason, a static (not public) restore breaks image links.  I can't see a difference in content (based on a diff of the static buckets from prod vs dev), so I'm not sure what is happening with this. 

Running tome to see if that restores images.

Update: After tome run it looks like images are available, but not logos?

arpage (commented):
Hi @jacobaaronyeager - I would recommend `--only-show-errors` for recursive `aws s3 cp` and `aws s3 rm` commands.

The amount of output is huge, and omitting it may speed things up a bit.

Original backup scripts use:

```
echo "Deleting s3 static site snapshot '$DEPLOY_TAG'. This could take a while."
aws s3 rm --only-show-errors --recursive s3://$S3_BUCKET/$BUCKET_BACKUP_FOLDER/$DEPLOY_TAG
```

and 
```
aws s3 cp --only-show-errors s3://$S3_BUCKET/$BUCKET_SOURCE_FOLDER/ s3://$S3_BUCKET/$BUCKET_BACKUP_FOLDER/$DEPLOY_TAG/ --recursive
```

arpage:
@jacobaaronyeager - While testing deletes, I planned to delete `TEST-dev-14913-2025-12-12--acl-test-0` but I don't see a command for removing a named snapshot.  Help for `clean` seems to only work on date ranges or everything.

jacobaaronyeager:
> @jacobaaronyeager - While testing deletes, I planned to delete `TEST-dev-14913-2025-12-12--acl-test-0` but I don't see a command for removing a named snapshot. Help for `clean` seems to only work on date ranges or everything.

Now why didn't I think of that???

arpage:
@jacobaaronyeager - Amy had mentioned incorporating a mechanism to limit the number of tome-spawned backups.   I noticed this because tome is running faster than in the Before Times, and when I've been making test changes in dev, I end up with 2 or 3 backups per run.   I think if (when!) we have content editors making lists of changes throughout the day, the backup count could get pretty high.

arpage:
@jacobaaronyeager 
```
❯ scripts/snapshot/local-manager.sh list
scripts/snapshot/manager.sh: line 1375: syntax error: unexpected "(" (expecting "}")
```

arpage:
Hi @jacobaaronyeager 

I tested all 3 backup/restore types successfully, 3 times.

Only once real concern - when doing restores, the process does not take into account the current state of the tome-disabled flag: i.e. if tome is already disabled, it will be re-enabled when the backup process is complete.  IMO, the backup processes should record the current state of both maintenance-mode and tome-disabled prior to starting, and then restore the saved state after completion.

A small nitpick is that one of the y/n prompts for (can't remember if it's a delete or restore) requires "yes/no" rather than "y/n".

Otherwise, works well and looks good :) !

jacobaaronyeager:
> Hi @jacobaaronyeager
> 
> I tested all 3 backup/restore types successfully, 3 times.
> 
> Only once real concern - when doing restores, the process does not take into account the current state of the tome-disabled flag: i.e. if tome is already disabled, it will be re-enabled when the backup process is complete. IMO, the backup processes should record the current state of both maintenance-mode and tome-disabled prior to starting, and then restore the saved state after completion.
> 
> A small nitpick is that one of the y/n prompts for (can't remember if it's a delete or restore) requires "yes/no" rather than "y/n".
> 
> Otherwise, works well and looks good :) !

Changes implimented!

.docker/Dockerfile-cms
```diff
@@ -258,6 +258,7 @@ COPY .docker/motd /etc/motd
 COPY --chown=nginx:nginx composer.json /var/www/composer.json
 COPY --chown=nginx:nginx composer.lock /var/www/composer.lock
 COPY --chown=nginx:nginx scripts /var/www/scripts
+COPY --chown=nginx:nginx bin /var/www/bin
```
akf: Note: don't deploy bin to the cms.

akf: Okay, I see why we need snapshot-backups there.

.docker/Dockerfile-cms
```diff
@@ -258,6 +258,7 @@ COPY .docker/motd /etc/motd
 COPY --chown=nginx:nginx composer.json /var/www/composer.json
 COPY --chown=nginx:nginx composer.lock /var/www/composer.lock
 COPY --chown=nginx:nginx scripts /var/www/scripts
+COPY --chown=nginx:nginx bin/snapshot-backups /var/www/bin/snapshot-backups
```
akf: I believe you can take this line out. I have concluded that we don't need these scripts deployed to the cms; nothing in the scripts under `scripts` appears to call them, and they couldn't work there as things currently stand for two reasons I can see: 
1. We don't install the `cf` cli on the cms, and it's needed for these.
2. These scripts call other `bin` scripts that won't be available, like `bin/cloudgov/get-s3-access` and `bin/deploy/includes`.

scripts/snapshot/test.sh
akf, line 1: Running `scripts/snapshot/test.sh` on dev, I see a couple of things: 
```
📅 Testing date calculation compatibility...
Date command info: Mon Jun 15 21:13:46 UTC 2026
⚠️ Advanced date calculations not available
Backup cleanup will be disabled in this environment
```
Will that be a problem?  It seems as if we want that to work. Is this just wrong? I ran `./scripts/snapshot/manager.sh clean static,public` (which is what `tome-sync.sh` would run) and it seems to do the job. 

And then this, at the end, just looks like the test script is expecting something you've since changed: 
```
❌ Missing backup all command call
❌ FAILED: Backup Integration in tome-sync.sh
```

bin/snapshot-backups/public-snapshot-list
akf, line 1: Please change the shebang lines on these to `#!/bin/bash`

Why: The default shell on ubuntu, `dash`, is extremely limited and I get an error message when I run these. (The error message actually comes from `bin/deploy/includes`, which does specify `bash`, but that doesn't help when a file is sourced.) I suspect that these scripts also wouldn't work on alpine, although without the `includes` script, `public-snapshot-list` simply fails silently.

Here's what the error message looks like, just FYI: 
```
~/dev/usagov-2021 % bin/snapshot-backups/public-snapshot-list dev --with-dates
bin/snapshot-backups/public-snapshot-list: 10: bin/snapshot-backups/../deploy/includes: Syntax error: "(" unexpected
```

scripts/snapshot/backup-system.conf
```diff
@@ -0,0 +1,68 @@
...
+ENABLE_DB_BACKUPS=true
+
+# Enable/disable automatic cleanup of old database backups (true/false)
+ENABLE_DB_AUTO_CLEANUP=true
+
+# Time to run daily database backups (24-hour format, EST)
+DB_BACKUP_TIME="21:00"
+
+# Number of days to retain database backups (SSPP minimum: 180)
+DB_BACKUP_RETENTION_DAYS=180
```
akf, line 65: Confirmed 180 days is a good number. We don't actually have a directive to save backups for a specific time period, but we've settled on 6 months elsewhere.

scripts/snapshot/local-manager.sh
akf, line 1: Please also change the shebang to `#!/bin/bash` here. Without that change I get: 
```
~/dev/usagov-2021 % ./scripts/snapshot/local-manager.sh --help
./scripts/snapshot/local-manager.sh: 162: Syntax error: redirection unexpected
```

What would you think of moving this to the `bin/` directory, since it's meant to be local only? Then it won't even be installed on the cms. I know conceptually it's nice to have it right next to manager.sh. Maybe just a comment in this script about it's being a local version of `scripts/snapshot/manager.sh` would do?

scripts/snapshot/setup-cron.sh
```diff
@@ -0,0 +1,203 @@
...
+    echo "Usage: $0 [command]"
+    echo ""
+    echo "Commands:"
+    echo "  setup       Setup database backup cron job"
+    echo "  remove      Remove backup cron jobs"
+    echo "  status      Show current cron jobs (default)"
+    echo "  test        Test the exact cron command (simulates cron environment)"
+    echo ""
+    echo "Configuration:"
+    echo "  DB_BACKUP_TIME: Set backup time in EST (format: HH:MM)"
```
akf: I have opinions about timezones. EST is only a thing for part of the year, and I was about to say you should just say ET. But what you're doing is setting backup time in UTC-5. So you could say "Set backup time in UTC-5 (EST)" and I at least would be less confused. 

But why use Eastern time at all, when the server is doing everything in UTC? It's not a big deal to look up the time conversion and make sure you're telling the computer what you want, plus we're pretty much going to set this once and never run it interactively. When there's a conversion like this going on I feel that I have to look at the code to be sure I understand what it's going to do ... and in this case it actually wasn't what the comment made me think it would do. 

**What time should it be?**
Part of the reason we delay deployments until 8pm ET is that the Contact Center is still active until 8pm. So although I said something different the other day, I think we should go with `01:00 UTC`, which will be 8pm EST or 9 pm EDT.

scripts/tome-sync.sh
```diff
@@ -174,12 +174,14 @@ LCF=0
 echo "Lower-casing files:"
 old_IFS="$IFS"
 IFS=$'\n'
-for f in `find $RENDER_DIR/*`; do
+# Process paths in reverse depth order (deepest first) to avoid "No such file or directory" errors
+# when parent directories are renamed before their children
+for f in `find $RENDER_DIR/* -depth`; do
```
akf, line 179: 👍

scripts/tome-sync.sh
```diff
@@ -430,6 +432,41 @@ if [ "$TOME_PUSH_NEW_CONTENT" == "1" ]; then
...
+          echo "Starting automatic backups..." | tee -a $TOMELOG
+
+          # Create static and public backups using manager.sh backup command
+          # The manager.sh script will handle all the logic, config loading, and smart detection
+          # Run from /var/www to ensure manager.sh can find its dependencies
+          # Use --throttle to skip backups if one was created recently (configurable via BACKUP_THROTTLE_HOURS)
+          if (cd /var/www && $BACKUP_MANAGER backup static,public --throttle) 2>&1 | tee -a $TOMELOG; then
+              echo "Automatic backup completed successfully." | tee -a $TOMELOG
+          else
+              echo "Warning: Backup process encountered issues." | tee -a $TOMELOG
```
akf: Consider giving warnings about backup issues a more distinctive pattern so we can easily find them in New Relic (either for a dashboard or for creating alerts). For example, we use this query for "tome errors": 
```
SELECT tags.space_name , message FROM Log WHERE message LIKE '%WARNING: ***%***%'
```

Even though these warnings are in the tome-sync script, I don't think they are "tome errors." So I'll suggest a different pattern -- '%WARNING (backup): ***%***%', which would make the above message look like: 
```
   echo "WARNING (backup): *** Backup process encountered issues. ***"
```

scripts/tome-sync.sh
```diff
@@ -430,6 +432,41 @@ if [ "$TOME_PUSH_NEW_CONTENT" == "1" ]; then
...
+          fi
+
+          # Run cleanup using manager.sh clean command
+          # Clean static/public backups older than BACKUP_RETENTION_DAYS (default: 180 days, SSPP requirement)
+          # Use --non-interactive flag to skip confirmation prompts in automated context
+          CLEANUP_DAYS=180
+          if [ -n "$BACKUP_RETENTION_DAYS" ] && [ "$BACKUP_RETENTION_DAYS" -ge 180 ]; then
+              CLEANUP_DAYS=$BACKUP_RETENTION_DAYS
+          elif [ -n "$BACKUP_RETENTION_DAYS" ] && [ "$BACKUP_RETENTION_DAYS" -lt 180 ]; then
+              echo "Warning: BACKUP_RETENTION_DAYS ($BACKUP_RETENTION_DAYS) is less than SSPP minimum of 180 days. Using 180 days." | tee -a $TOMELOG
```
akf: Strike "SSPP" here. The SSPP says 180 days for logs (and might have conflicting info elsewhere), but it doesn't specify this for backups.

akf (changes requested):
I added comments to the files; here are the highlights: 
 - I think I've confirmed that nothing you've set up to run in the cms needs the scripts in `bin/snapshot-backups`, so you don't need to add those to the cms build. 
 - Please change the shebang line for local scripts to `#!/bin/bash` (explanation in comments). By contrast, the cms won't have `/bin/bash`, so keep the scripts in `scripts` as is (except for `local-manager.sh`)
 - I'm not sure the cleanup is running correctly when `tome-sync.sh` calls it. It doesn't produce any error messages, but I don't think it deleted the backups. I ran it "by hand" (see comment) and it did work, which surprised me a little because `test.sh` said it would be disabled. I canceled that job when it was almost finished deleting static backups but hadn't gotten into the public ones, so there are still old backups to test against. 
 - I think you should move `local-manager.sh` to the `/bin/snapshot-backups` directory, because it's meant to be run locally.
 - I think `setup-cron.sh` should just use UTC and skip the timezone conversions. (Timezone rant redacted.) 
 
I only skimmed the README this time. There will probably be some tweaks to go along with any changes you make in this round.

scripts/snapshot/setup-cron.sh
akf: Per your comment that 1:00 UTC is within the deployment window, let's go with 23:00 UTC so a backup happens *before* deployment.

scripts/snapshot/setup-cron.sh
```diff
@@ -0,0 +1,196 @@
...
+    # Determine working directory for cron job
+    CRON_WORK_DIR="/var/www"
+    if [ -n "$VCAP_APPLICATION" ] && [ -d "/var/www" ]; then
+        CRON_WORK_DIR="/var/www"
+    else
+        CRON_WORK_DIR="$PROJECT_ROOT"
+    fi
+
+    # Add new cron job (using new command format)
+    (crontab -l 2>/dev/null; echo "$minute $utc_hour * * * cd $CRON_WORK_DIR && $BACKUP_DIR/manager.sh backup db >/dev/null 2>&1") | crontab -
```
akf: $utc_hour isn't defined here, so I think this will just fail (I didn't test it).

scripts/snapshot/setup-cron.sh
```diff
@@ -0,0 +1,196 @@
...
+    print_status $GREEN "Configuration:"
+    echo "  Database backups enabled: $ENABLE_DB_BACKUPS"
+    echo "  Backup time: $DB_BACKUP_TIME EST"
+}
+
+# Test the cron backup command in a simulated cron environment
+# Executes the exact command that cron will run to verify functionality
+test_cron_command() {
+    print_status $YELLOW "Testing cron command execution..."
+    print_status $BLUE "This simulates thUTCxact environment and command that cron will use"
```
akf: Typo here -- "the exact" changed to "thUTCxact"

akf (commented):
I see some small things: 

1. scripts/snapshot/test.sh still ends with complaints. I'd hate to push this through with a failing test. (It looks like the test just needs to be updated to match the current reality): 
```
❌ Missing backup all command call
❌ FAILED: Backup Integration in tome-sync.sh
```

2. I think I caught a misnamed variable that will prevent `setup_cron.sh` from working; see comment. (I did not test to be sure.) Also caught a typo in a comment in that script.

akf (approved):
My comments are addressed and so I'll approve. I have not actually re-tested running a backup.

arpage (approved):
👍

PR #2512, opened 2025-12-05
USAGOV-1685 - site banner config_pages implementation

arpage:
## Jira Task

## Description

Addition of config for new implementation of Site Banners

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [x] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions

1. Go to:  /admin/structure/config_pages
2. There should be one entry, "Site Banner" - click on Edit
3. Here you have the ability to add/remove paragraphs, each of which will be one banner, in case more than one are needed.
4. If you enter text for the Title field, the banner should be displayed after the next tome run.
5. To stop banners displaying, remove the Title/Body text. Optionally the paragraph(s) can be deleted.

### Change Requirements

- [x] Requires New Config
- [ ] Requires New Content

### Validation Steps

1. When adding banner(s): Ensure the banner(s) appear after tome has regenerated the static site.
2. When removing banner(s): Ensure the banner(s) disappear after tome has regenerated the static site.

## Security Review

- [x] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

6. Find the commit of this pull request.
7. Build and deploy the changes.
8. Update the Jira ticket by changing the ticket status to `Review in Test` and add a comment. State whether the change is already visible on [cms-dev.usa.gov](http://cms-dev.usa.gov/) and [beta-dev.usa.gov](http://beta-dev.usa.gov/), or if the deployment is still in process.


config/sync/core.extension.yml
```diff
@@ -107,6 +108,7 @@ module:
   usagov_blog_date_menu: 0
   usagov_ckeditor5_source_editing_fixup: 0
   usagov_directories: 0
+  usagov_extra_functions: 0
```
akf, line 111: It looks like usagov_extra_functions just has an update hook to remove site_banner block content entities. Is this redundant (or partially redundant) with the `remove-existing-site-banner-infrastructure.php` script linked from Jira? It looks like it's partially redundant.

docs/ConfigPagesSiteBanner.md
akf, line 1: This was very helpful. I was scratching my head over where the banner was actually getting inserted into the page. 

I'm still scratching my head over why we don't have the static-site problem with embedded paragraphs elsewhere! Are you sure the _preprocess routines aren't needed for the CMS display as well? (Maybe the difference is that other paragraphs are within a block?)

docs/ConfigPagesSiteBanner.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Benefits of This Approach
+
+1. **Centralized Management**: All site banners managed in one place via Config Pages
+2. **Multiple Banner Support**: Can display multiple alert banners simultaneously
+3. **Analytics Tracking**: Each banner includes the `data-analytics="site-banner"` attribute
+4. **Maintainability**: Uses the existing USWDS alert paragraph system
+5. **Flexibility**: Supports all USWDS alert types and configurations
+6. **Accessibility**: Maintains all accessibility features from the original implementation
+7. **Language Support**: Works with both English and Spanish content
```
akf: I am not seeing the language support -- I get the same banner on all pages! But we'll just need to fix that.

web/modules/custom/usa_twig_vars/usa_twig_vars.module
```diff
@@ -278,36 +274,46 @@ function usa_twig_vars_preprocess_region(array &$variables): void {
...
+      foreach ($config_pages_site_banners as $config_page) {
+        // Check language matches current site language or is undefined.
+        // Config_pages can be language-neutral (und), which should appear on all language variants.
+        $config_page_lang = $config_page->language()->getId();
+        if ($config_page_lang !== $site_language && $config_page_lang !== 'und') {
+          continue;
         }
 
+        // Check if the config page has alert field with content.
+        if ($config_page->hasField('field_alert') && !$config_page->get('field_alert')->isEmpty()) {
```
akf, line 309: PHPStan has some issues with `hasField` and `get` on line 292 here; says `Drupal\Core\Entity\EntityInterface` doesn't have those methods. Obviously this is working, so I suspect either this code isn't exposing `$config_page`'s true nature, or our PHPStan setup is borked somehow. 

There are similar complaints on lines 294, 393, and 394. This isn't a "must fix" but one of us should at least take a look at it. We should do another round of cleanup of PHPStan errors and make that scan blocking (like Lint) soon.

akf (changes requested):
This is working well. I do see a few things that need to be adjusted: 

1. We need English and Spanish versions of the banner. I don't think that means we need a language checkbox; they can stay "und", I think it means separate "Site Banner English" and "Site Banner Spanish" config pages, and inserting the appropriate one in the header based on the language. 
2. There's a gap above the banner that wasn't present in the banners you put up on October. My understanding is that it should slam right up against the usa banner, and if there are multiple site banners they should slam up against each other without gaps. 
3. Content Administrators should be able to manage site alerts, and it turns out they cannot. I added the "Edit All ConfigPages entities" on my local site and what it gives them seems perfect. 
4. The banners need to be added to the Wizard pages too. There are twig templates under `wizard_2024`: `region-header-top-wizard.html.twig`. Looks like there's some `site_banners` code to remove there too. (As a reminder, those are used for the Scams wizard. I think they wound up with their own templates because they're implemented as taxonomy.) 
5. Maybe that applies to the blog too, since it's a view? 

Also I made some comments inline.

config/sync/core.extension.yml
jacobaaronyeager: I'm not certain.  The update hook was supposed to remove the old block stuff on it's own.  Perhaps @arpage should weigh in?

arpage: @akf @jacobaaronyeager 
To test again, I will restore tonight's prod backup to dr, with the site-banners branch deployed. I'll see what happens during bootstrap and report back.

web/modules/custom/usa_twig_vars/usa_twig_vars.module
jacobaaronyeager: I pushed something to phpstan.neon that will hopefully help phpstans discoverability, but no guarantees.  I'm 98% certain it's an issue with phpstan not really understand Drupal's inheritance structure, so I think that's all that really needs to be done.

config/sync/core.extension.yml
arpage: @jacobaaronyeager - if I run the php script attached to the ticket on a fresh db restore from prod, then deploy the site banner redux branch - everything works as far as the new config getting added, and the new banner page being present/working. 👍

akf: Okay, I'm still uncertain whether the usagov_extra_functions module is needed (or whether the script is needed if usagov_extra_functions is in place), but I've decided it doesn't matter.

web/modules/custom/usa_twig_vars/usa_twig_vars.module
```diff
@@ -278,36 +274,63 @@ function usa_twig_vars_preprocess_region(array &$variables): void {
...
+          foreach ($context_data as $context_item) {
+            if (is_array($context_item) && isset($context_item['language'])) {
+              $context_language = $context_item['language'];
+              break;
+            }
           }
         }
 
+        // Skip if context language doesn't match current language.
+        // Allow config_pages with no language context (NULL) to appear on all pages.
```
akf, line 303: Let's remove this comment and the `$context_language !== NULL && ` from the next line. That is, let's not display banners that lack a language context.

docs/ConfigPagesSiteBanner.md
```diff
@@ -0,0 +1,168 @@
...
+- **Pre-render paragraph entities using the paragraph view builder** for static site compatibility
+- Extract multiple referenced alert paragraphs from the config_pages field_alert field
+- Add pre-rendered config_pages site banners to the `config_page_alerts` array for template rendering
+- Support multiple paragraphs displaying as separate banners
+
+**Language Handling:**
+- Config_pages uses a **context field** (not entity langcode) to store language-specific data
+- When language context is enabled, the context field contains serialized data like: `a:1:{i:0;a:1:{s:8:"language";s:2:"en";}}`
+- The preprocess function extracts and checks the language from the context field
+- Config_pages with `NULL` context language appear on all language variants
```
akf: Change "appear on all language variants" to "are not displayed (editors should not be able to create a config_page without a language context and will be unable to manage them)" 
... and put that after the first two bullets.

akf (changes requested):
I've re-reviewed and just have the one request we talked about yesterday -- let's not display banners with NULL language context, should one somehow get into the system. The admin UI, as it's configured, doesn't enable content editors to manage site banners that lack a language context so it would be a problem for them if one started showing up for some reason. (And it's a small change.) 

I have been noticing a caching problem on the English home page and I think I should point it out here: The first time I add a site banner, it doesn't show up there right away. So when I tested this time, I carefully avoided loading any pages after I added site banners (one for each language context), and ran tome. In the tome output, `index.html` lacked the banner, while `es/index.html` had it. All the other pages seemed to have it, though. 

Then I went back to my web browser and loaded the home page both as the logged-in content editor and in an incognito window, and the banners appeared. Ran tome again and banners appeared as desired. After that, I was unable to reproduce the problem by deleting and re-adding banner content; it seems to occur only the first time I create a banner after updating to this branch. Weird!

jacobaaronyeager:
> I've re-reviewed and just have the one request we talked about yesterday -- let's not display banners with NULL language context, should one somehow get into the system. The admin UI, as it's configured, doesn't enable content editors to manage site banners that lack a language context so it would be a problem for them if one started showing up for some reason. (And it's a small change.)
> 
> I have been noticing a caching problem on the English home page and I think I should point it out here: The first time I add a site banner, it doesn't show up there right away. So when I tested this time, I carefully avoided loading any pages after I added site banners (one for each language context), and ran tome. In the tome output, `index.html` lacked the banner, while `es/index.html` had it. All the other pages seemed to have it, though.
> 
> Then I went back to my web browser and loaded the home page both as the logged-in content editor and in an incognito window, and the banners appeared. Ran tome again and banners appeared as desired. After that, I was unable to reproduce the problem by deleting and re-adding banner content; it seems to occur only the first time I create a banner after updating to this branch. Weird!

I know why this banner problem is.  Because config_pages do not have URIs themselves, they are not normally discoverable by Tome.  I got around this by having them embed into whatever page they were being used on and Tome would pick them up that way.  I can see that if you never load the page, Tome might pick up a cached version of that page without the config_page embedded, though it would get it the second time through since it would render it by running.

akf (approved):
All my concerns have been addressed!

PR #2572, opened 2026-02-04
USAGOV-2616: deploy to prod

akf:
## Jira Task

https://cm-jira.usa.gov/browse/USAGOV-2616


akf:
For anyone looking at this and puzzled by the "failed" jobs, they're carried over from the stage deployment. I canceled the jobs we didn't run in an attempt to clean up the view in CircleCI. It turns out that produces a more confusing result than just leaving them hanging.

PR #2651, opened 2026-03-25
Usagov 2679 update bf node dependancies

jacobaaronyeager:
## Jira Task

https://gsa-standard.atlassian-us-gov-mod.net/browse/USAGOV-2679

## Description
Brings current all Benefit Finder dependencies--sotrybook, vite, cypress, node, react, etc.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [X] Frontend (Twig, Sass, JS)
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions
Deploy and fully regression test the Benefit Finder, and/or run Cypress tests following instructions in web/modules/custom/usagov_benefit_finder/app-source/cypress/README.md

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


web/modules/custom/usagov_benefit_finder/modules/usagov_benefit_finder_app/usagov_benefit_finder_page/js/benefit-finder.min.js
akf, line 1: It seems weird to me that this file is being deleted. And ever since I noticed that, this app has stopped working for me! I think I was getting a cached version. 

So I see that this file gets re-generated during the build process for the benefit finder app. That happens for me, but I get a syntax error when I load a page: 

```
Uncaught SyntaxError: Cannot use 'import.meta' outside a module (at benefit-finder.min.js?v=1.x:11:11985)
```

But what we want here is to check this file into git. We don't want to re-build it during deployment (and in fact as far as I know we won't). It should be an asset that basically comes along with the module.

.gitignore
```diff
@@ -128,6 +128,9 @@ web/modules/custom/usagov_benefit_finder/app-source/dist/
 web/modules/custom/usagov_benefit_finder/app-source/.env.local
 web/modules/custom/usagov_benefit_finder/app-source/coverage/
 web/modules/custom/usagov_benefit_finder/app-source/storybook-static/
+web/modules/custom/usagov_benefit_finder/modules/usagov_benefit_finder_app/usagov_benefit_finder_page/assets/
```
akf: Again, these files should be committed. Just the ones under app-source should not. 

Remember that originally, the contents of app-source were part of a separate repo that was used to build this module.

sources/benefit-finder/src/Routes/ResultsView/__tests__/__snapshots__/index.spec.jsx.snap
akf, line 1: The sense I'm getting from these changes is that you made them fit the data you have.  We will need a database snapshot (or some other way of generating test data) that matches the tests, I think. If it's a database snapshot we can just keep it in Google Drive and note its location in the README. It can be a snapshot from your local dev if that's a good thing to test against. 

BUT ... 

I'm concerned about the batch that starts at line 3516, because it looks like a lot of content was deleted and ... replaced with empty divs? That doesn't seem right, and I wonder whether what happened is that these tests pick the first benefit-finder option on a page and you wound up with it picking a set of criteria that didn't return useful data to test.

web/modules/custom/usagov_benefit_finder/app-source/cypress/README.md
```diff
@@ -0,0 +1,94 @@
...
+**Step 1:** Build the Storybook static site
+
+```bash
+npm run cy:build:storybook
+```
+
+**Step 2:** Start the Storybook server and run all tests
+
+```bash
+npm run cy:run:pipeline
```
akf: I don't think this works for me; it seems to just hang.

web/modules/custom/usagov_benefit_finder/app-source/cypress/README.md
```diff
@@ -0,0 +1,94 @@
...
+```
+
+This uses `concurrently` to spin up `http-server` on port 6006 serving `./storybook-static`, then runs Cypress headlessly in Chrome. The server is killed when the test run finishes.
+
+---
+
+### Option 2 — Run tests against an already-running Storybook
+
+If Storybook is already running (e.g. via `npm run cy:dev:storybook`), you can run the tests directly:
```
akf: I see an warning about "Failed to load preset" due to an error "Cannot find package 'typescript'" when I build or run storybook. It does seem to run, though, so perhaps it's not a blocking problem.

sources/benefit-finder/cypress/README.md
```diff
@@ -0,0 +1,94 @@
...
+This uses `concurrently` to spin up `http-server` on port 6006 serving `./storybook-static`, then runs Cypress headlessly in Chrome. The server is killed when the test run finishes.
+
+---
+
+### Option 2 — Run tests against an already-running Storybook
+
+If Storybook is already running (e.g. via `npm run cy:dev:storybook`), you can run the tests directly:
+
+```bash
+npm run cy:run:e2e
```
akf, line 64: This works!

sources/benefit-finder/cypress/README.md
```diff
@@ -0,0 +1,94 @@
...
+---
+
+## Configuration
+
+- **Default config:** `cypress.config.js` — targets `http://localhost:6006`, runs in Chrome with 2 retries on failure
+- **Production config:** `cypress.prod.links.config.js` — used for link-checking against the live site (`npm run cy:run:prod:links:e2e`)
+
+---
+
+## CI
```
akf, line 128: We don't have a CI pipeline to run these scripts, right?

akf (changes requested):
A lot of this looks good but I ran into some snags; one is a show-stopper.

**Must fix:** we need to add some of the deleted files back and remove some of the new gitignore lines -- see my comments. We shouldn't be deleting or ignoring anything that's not under the app_source directory, as far as I can tell. We don't want to re-build those except when someone is updating things in app_source. 

Annnddd, I see that we've been doing that in Dockerfile-cms and I guess I missed that before and thought we were just doing it for local dev. I think we should unwind that! Obviously it was working before, even for me, and now it does not work for me ... but also we really should generate these assets, check them in, and let them be static until we need to change something. 

**Must check:** It turns out the benefit-finder.min.js that built on my system doesn't work. I don't know why but I captured the error. I missed this at first; apparently I had the old version of the file cached for localhost. So make sure you're testing the finished product after clearing both drupal's cache and your web browser's. 

**Maybe an issue:** but I'm inclined to ignore ... two of the cypress tests failed for me. I don't know whether they worked before, though.

sources/benefit-finder/src/Routes/ResultsView/__tests__/__snapshots__/index.spec.jsx.snap
jacobaaronyeager: This test used mock data, not anything from the DB.  The issue was the mismatch between Node 24 that's now used in BF and Node 22 that the tests are expecting.  The larger issue is that right now these tests output a big old blob and say if it's good or not; we could instead target based on our actual tests and the output would only show what's appropriate.

akf (changes requested):
This still isn't building right for me, apparently. When I open a benefit-finder page, the tool doesn't load and I get this error message in my console: 

```
benefit-finder.min.js:11 Uncaught SyntaxError: Cannot use 'import.meta' outside a module
```
I still think we should reinstate the benefit-finder.min.js file (I see  you removed it from .gitignore) and rearrange the build so that it happens only for local dev. Then we can rebuild locally, check in the js, css, and assets, and they'll be deployed as-is. (Handwaving here because I don't have a simple plan for that, but it might be out of scope for this task anyway.)

docs/CODEOWNERS
akf, line 1: I approve of deleting this!

sources/benefit-finder/README.md
akf, line 1: This file has the following instructions, which don't entirely work for me: 

> Manual work in the source directory:
>
> cd sources/benefit-finder
> npm install
> 
> # Development mode with hot reload
> npm start
> 
> # Build committed Drupal module assets
> npm run build
> 
> # Run tests
> npm test

Specifically, `npm start` starts a server on port 3000 and opens localhost:3000/death in a browser, but it's trying to load a json file that doesn't exist (sound familiar?) -- I get this log: 

```
 VITE v8.0.2  ready in 243 ms

  ➜  Local:   http://localhost:3000/
  ➜  Network: http://192.168.1.7:3000/
  ➜  Network: http://100.64.0.1:3000/
  ➜  press h + enter to show help
4:21:59 PM [vite] http proxy error: /s3/files/benefit-finder/api/life-event/death.json
Error: Must set target or forward
    at ProxyServer.web (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:17583:43)
    at viteProxyMiddleware (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:17840:10)
    at call (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:6166:5)
    at next (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:6119:4)
    at viteCachedTransformMiddleware (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:24912:3)
    at call (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:6166:5)
    at next (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:6119:4)
    at hostValidationMiddleware$1 (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:16064:3)
    at call (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:6166:5)
    at next (file:///home/amykfarrell@ent.ds.gsa.gov/dev/usagov-2021/sources/benefit-finder/node_modules/vite/dist/node/chunks/node.js:6119:4)

```

scripts/benefit-finder/sync-usagov-theme.sh
akf, line 1: sync-usagov-theme.sh isn't mentioned in the readme and probably should be? I see that it's hooked in to some other commands in sources/benefit-finder/package.json, so maybe it's not a step I need to do at any point.

sources/benefit-finder/cypress/README.md
akf, line 1: After re-generating benefits-finder.min.js, I'm getting a single failure from the dataLayer.cy.js test `results page with eligible benefits has a bf_page_change and bf_count events`. Do you get that too? 

We probably don't need to fix that in this PR, but it would be interesting if you get a different set of results. It might mean instructions are missing, or that I missed following some instructions.

akf, line 1: In the Static Storybook workflow, all tests pass for me.

sources/benefit-finder/cypress/README.md
```diff
@@ -0,0 +1,130 @@
...
+| `selected-criteria-eligibility-benefits.cy.js` | Eligibility results for selected criteria |
+
+---
+
+## Configuration
+
+- **Default config:** `cypress.config.js` targets `http://localhost:6006`, retries failed runs twice in headless mode, and excludes `cypress/e2e/usagov-public-site/*.cy.js`
+- **Production links config:** `cypress.prod.links.config.js` targets `https://www.usa.gov` and is wired to `cypress/e2e/usagov-public-site/links.cy.js`
+
+At the time of this update, the checked-in Storybook suite lives under `cypress/e2e/storybook/`. If you need the production links workflow, confirm that the `usagov-public-site` spec file exists in your branch before using `npm run cy:run:prod:links:e2e`.
```
akf: What is this `usagov-public-site` file?

sources/benefit-finder/cypress/README.md
akf: There are some details here that might be helpful if we were to set up a pipeline, but I think we should comment this stuff out for now.

web/modules/custom/usagov_benefit_finder/modules/usagov_benefit_finder_app/usagov_benefit_finder_page/usagov_benefit_finder_page.libraries.yml
```diff
@@ -1,7 +1,9 @@
 benefit_finder_app:
   version: 1.x
   js:
-    js/benefit-finder.min.js: {}
+    js/benefit-finder.min.js:
+      attributes:
+        type: module
```
akf, line 6: TIL

akf (changes requested):
**Again I have a must-fix:**

This builds a react app with some text that is different from what we have in production. I can call out one specific case: the first screen for each of the "life events" should have a heading "Steps to find benefits". This builds "Steps to find benefit". That is due to a change in `sources/benefit-finder/src/shared/locales/en/en.json` -- happily, this is also my favorite file to edit to quickly see whether the "watch changes" functionality is working. 

I do see the benefits/benefit change in this PR. I suspect you were bringing the .json file in line with the cypress tests and probably should have made the change in the tests. 

Also, now that I've changed that `en.json` file and changed it back, git shows the `benefit-finder.min.js` file in `modules/usagov_benefit_finder_app/usagov_benefit_finder_page/js`  as modified. I also got more errors from the cypress tests the first time I ran them, and I think that's why. (Now, when I restore benefit-finder.min.js from git, it immediately gets updated by the watcher. Must be a timestamp thing that it didn't happen the first time.) 
**So please make sure the built files that are committed match up with the source that creates them.**


**The rest of these are non-blocking comments** -- please look through them and see if there's anything you think is worth addressing.

akf (changes requested):
My "must fix" from the last round is resolved.  Now there is a CSS problem, and another thing that worries me. 

Worry first: I wind up with some untracked files -- that is, the benefit-finder build must be creating them and they're not checked in. The first is a whole directory of assets; the second is a minified css file: 
```
web/modules/custom/usagov_benefit_finder/modules/usagov_benefit_finder_app/usagov_benefit_finder_page/assets/
web/modules/custom/usagov_benefit_finder/modules/usagov_benefit_finder_app/usagov_benefit_finder_page/css/benefit-finder2.min.css
```

The CSS problem is that the "Before you start:" heading on the first page of the tool has `width: 100px` applied. Web inspector tells me three things: 
 - The selector this is applied to is `#benefit-finder .bf-intro .bf-intro-process-notices .bf-intro-process-notices-heading`
 - It is coming from benefit-finder-min.css (which is what I'd expect, but that benefit-finder2.min.css file was generated, so just confirming!)
 - On prod, the same selector has `width:100` and this is called out by inspector as invalid and ignored. So I think a syntax error got fixed somewhere but what actually needs to happen is that `width` needs to be removed entirely somewhere.

akf (approved):
It builds/rebuilds as expected and as far as I can tell, produces the same screens as what we see in production.

PR #2635, opened 2026-03-17
USAGOV-2668: Update drupal core and dependencies

arpage:
## Jira Task

https://gsa-standard.atlassian-us-gov-mod.net/browse/USAGOV-2668

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Frontend (Twig, Sass, JS)
  - Add screenshot showing what it should look like
- [ ] Drupal Config (requires "drush cim")
- [ ] New Modules (requires rebuild)
- [ ] Documentation
- [ ] Infrastructure
  - [x] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

## Testing Instructions

I tested the basic CMS functionality (CRUD and revisions).  I did not test media or ckeditor in any depth.

### Change Requirements

- [ ] Requires New Documentation (Link: {})
- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

CMS functionality should be intact, and the static site should generate normally.

## Security Review

- [ ] Adds/updates software (including a library or Drupal module)
- [ ] Communication with external service
- [ ] Changes permissions or workflow
- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality
- Security review complete or not required

## Post PR Approval Instructions

Follow these steps as soon as you merge the new changes.

1. Go to the [USAGov Circle CI project](https://app.circleci.com/pipelines/github/usagov/usagov-2021).
2. Find the commit of this pull request.
3. Build and deploy the changes.
4. Update the Jira ticket by changing the ticket status to `Review in Test` and add a comment. State whether the change is already visible on [cms-dev.usa.gov](http://cms-dev.usa.gov/) and [beta-dev.usa.gov](http://beta-dev.usa.gov/), or if the deployment is still in process.

PR #2659, opened 2026-03-27
USAGOV-2668 - only packages required for update to drupal core 10.6.x

arpage:
## Jira Task

https://gsa-standard.atlassian-us-gov-mod.net/browse/USAGOV-2668

Recreating the PR, with _only_  the s3fs and drupal core updates and _without_ the autologout patch.  Also removed js_cookie from the extensions config.

**Update:**  Also removed unneeded and possibly buggy call to `$node->setPublished(FALSE);` - usage becomes deprecated w/ 10.6

**NOTE:** Here is the link to the original PR, which has some important reviewer notes:  https://github.com/usagov/usagov-2021/pull/2635

## Description

composer.json and composer.lock updated to apply new package versions.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [x] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

#### Setup
1. Shorten the autologout periods, so that we can test the autologout 2.0.2 quickly:
```
bin/drush config:set autologout.settings timeout 90 -y 
bin/drush config:set autologout.settings padding 10 -y
```
2. Open 2 new tabs to `localhost`, preferably in their own windows, because the timeout is short, and makes things difficult if you need to switch tabs
3. Run drush uli - and login on one of the tabs:
```
bin/drush uli
```
4. Refresh the other tab, so that you can see we are logged in, and have the admin UI present. 
5. In in this tab, open chrome console, paste the following and press enter:
```
(function keepServerSessionAlive() {
  // Clear the cross-tab activity flag so Tab A sees userIsActive() === false.
  localStorage.removeItem('autologout_page_activity');

  // Ping the server to keep autologout_last current (onRequest updates it).
  fetch(location.href, { credentials: 'same-origin', cache: 'no-store' })
    .then(() => console.log('[autologout test] server session refreshed, flag cleared:', new Date().toLocaleTimeString()));

  setTimeout(keepServerSessionAlive, 8000);
})();

console.log('[autologout test] Tab B running. Watch Tab A for logout.');
```

### Expected Results
1. Don't do anything else in either of the tabs - both tabs should say logged in beyond the 90 second timeout.

2. After the previous behavior is confirmed, refresh the second tab, so that the JS we pasted earlier is no longer active. Don't do anything else in either of the tabs - both tabs should log out after the 90 second timeout.

### Test CMS functionality
CMS functionality and A11Y regression tests should function as with the 10.5 core

## Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


arpage:
@jacobaaronyeager and @akf 

I fixed Issue #2 from the @jacobaaronyeager's comments in the [original PR](https://github.com/usagov/usagov-2021/pull/2635).

The other two issues can be dealt with separately, per @akf's additional comment on the [original PR](https://github.com/usagov/usagov-2021/pull/2635).

akf (changes requested):
I think we need to do a prove-it session on autologout. Your test isn't working for me the way you say it would. 

The first time I tried it, I saw the "Your session is about to expire" pop-up in both windows at the same time, which seems okay but isn't what you said to expect. 

Then I decided to clear drupal cache and my browser's cookies (well, I did a hard reload), and since then, it's not working at all. I get logged out without even seeing the dialog (in either window).

arpage:
> I think we need to do a prove-it session on autologout. Your test isn't working for me the way you say it would.

@akf - I'm seeing the same thing as you, when deployed to dr.  Testing local again.
If it's turns out it's working on local and not dr, is there a way to force regeneration of js/css?

akf:
> > I think we need to do a prove-it session on autologout. Your test isn't working for me the way you say it would.
> 
> @akf - I'm seeing the same thing as you, when deployed to dr. Testing local again. If it's turns out it's working on local and not dr, is there a way to force regeneration of js/css?

`drush cr` should regenerate the js/css -- rebuilding the js/css in the theme is a separate issue, but that isn't relevant here and it's also more automatic. You can also go into the "performance" settings and turn off aggregation of js, but `drush cr` is good for a one-time thing. 

I also did a hard reload in my browser after clearing the drupal cache, to eliminate any cached js there. 

Stepping back a little further, make sure you actually reinstalled the drupal packages when you switched to this branch. The `bin/drupal-update` script is what I used.

PR #2750, opened 2026-05-28
Harden benefits-finder PR workflow against shell injection

arpitjain099:
Following up on my earlier fixes here, this one tightens up `benfits-finder-test-pr.yml`.

The three browser test jobs (Chrome, Firefox, Edge) each run:

```yaml
- name: Checkout usagov-2021 working branch
  run: |
    cd usagov-2021
    git switch ${{ github.head_ref }}
```

`github.head_ref` comes from the source branch name of the incoming pull request, so a contributor controls it. When an expression like that is dropped straight into a `run:` block, the runner substitutes the raw value into the script before the shell executes it, so a crafted branch name could break out of the `git switch` and run arbitrary commands on the runner. This is the script-injection risk described in GitHub's hardening guide: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections

The change binds the value to a step-level `env:` variable and references the quoted shell variable instead:

```yaml
- name: Checkout usagov-2021 working branch
  env:
    HEAD_REF: ${{ github.head_ref }}
  run: |
    cd usagov-2021
    git switch "$HEAD_REF"
```

Passed through `env`, the branch name reaches the script as plain data rather than as part of the command text, so it can no longer be evaluated as code. The same fix is applied to all three jobs. Logic is otherwise untouched. Static analyzers such as CodeQL (`actions/expression-injection`) and zizmor flag the original form.


akf (approved):
Thanks for the write-up. It looks like GitHub moved their documentation around since you captured that link, by the way.

arpage:
@arpitjain099 - The change looks good.  Thank you for the fix, the detailed explanation, and the links for more information.  This is much appreciated.

cc: @akf

arpage (commented):
Looks good - thanks @arpitjain099

arpage (approved):

PR #2759, opened 2026-06-05
USAGOV-2687 update patch for autologout module

arpage:
## Jira Task

https://gsa-standard.atlassian-us-gov-mod.net/browse/USAGOV-2687

## Description

Fix issues present in the new 2.0.3 patch, with timeout handling in multiple tabs.
Incorporate the existing 2.0.3 patch, and layer our changes on top.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [x] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

### Ensure CMS auto logout still works:

- Deploy locally
- Login to the local cms via `drush uli` or similar
- Open another tab in a separate window, so you can see the current admin page, and another admin page (e.g. Content)
- Go to the admin modules listing, and search for *autologout*
- Confirm the version is 2.0.3, then click *Configure*
- Set each of the two timeout values (which are currently set to 1800 seconds) to 90 seconds, and save the config.
- Do nothing in either window/tab for 90 seconds, and you should see each window/tab show an Alert modal, asking if you want to reset the session.
- Click Yes, and the modal should disappear on the current window/tab
- Within 20-ish seconds the modal should disappear on the other window/tab
- Both windows/tabs should remain logged in, and allow you to perform admin tasks.
- After confirming this works, remain idle in both windows/tabs for another 90 seconds until you see the modal alerts again.
- This time don't click on either of them for 20 seconds
- After this interval, both windows/tabs should be logged out of the cms

### Testing Process Screenshots:

- Autologout config page: USAGOV-2687-test-001
- Single Tab Modal Alert: USAGOV-2687-test-002
- Multi-tab Modal Alert: USAGOV-2687-test-003

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


config/sync/autologout.settings.yml
```diff
@@ -24,3 +25,6 @@ disable_buttons: false
 yes_button: ''
 no_button: ''
 whitelisted_ip_addresses: ''
+cookie_secure: false
```
akf, line 28: Let's change this to "true". It doesn't affect localhost (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#block_access_to_your_cookies ), and we should be setting cookies to secure. 

It's a checkbox on the autologout configuration form, so I re-tested with that set and it works (locally of course). I see this in the patch as well, so there might be some nuance to making this change. 

I was wondering why we haven't been dinged for this and then remembered that we're only having un-authenticated scans run!

patches/drupal/autologout-multitab-timeout-fix.patch
```diff
@@ -0,0 +1,492 @@
...
+     'alt_logout_url' => _autologout_get_url_with_token('autologout.alt_logout'),
+     'ajax_logout_url' => _autologout_get_url_with_token('autologout.ajax_logout'),
+     'ajax_set_last_url' => _autologout_get_url_with_token('autologout.ajax_set_last'),
+diff --git a/config/install/autologout.settings.yml b/config/install/autologout.settings.yml
+index c6945da..fc5a1aa 100644
+--- a/config/install/autologout.settings.yml
++++ b/config/install/autologout.settings.yml
+@@ -22,3 +22,4 @@ use_watchdog: true
+ whitelisted_ip_addresses: ''
+ cookie_secure: false
```
akf, line 19: Again, change to true? I'm not sure how this is going to play out with the patch; definitely don't want the patch to overwrite our setting.

akf (approved):
This looks great and it works well for me. 

I'm requesting a change that I consider optional (but I will probably create a new ticket if we skip it this time) -- to change the cookie "cookie_secure" setting to "true". 

If that's more than a quick change I think we should merge this and resolve the autologout version issue sooner rather than later; that setting is essentially what we're getting with the old module.

PR #2805, opened 2026-07-16
USAGOV-2826: Range-check IPv4 octets in validate-cidr CIDR path

arpitjain099:
I work on supply-chain security for open-source projects and was reading through the deploy scripts.

## Description

`scripts/validate-cidr` builds the nginx `allow` list and is meant to warn on any malformed entry ("This should be investigated ASAP"). Two out-of-range IPv4 octets slip through when the entry is in CIDR form:

- `valid_ip()` splits the address on `.` only, so for `10.0.0.999/8` the last field is `999/8`. The `-gt 255` test forces bash arithmetic, which evaluates `999/8` as `124`, so the check passes.
- `valid_cidr()` (the fallback) never range-checks the octets at all, so `300.0.0.0/8` is accepted as an aligned network even though `300` is not a valid octet.

The plain forms `10.0.0.999` and `300.0.0.0` are already rejected correctly, so it is only the CIDR path that is missing the range check.

Before this change:

```
$ scripts/validate-cidr "300.0.0.0/8"
INFO: 1 IP addresses checked        # accepted, no warning, exit 0
$ scripts/validate-cidr "10.0.0.999/8"
INFO: 1 IP addresses checked        # accepted, no warning, exit 0
```

After:

```
$ scripts/validate-cidr "300.0.0.0/8"
IP VALIDATION WARNING: Encountered invalid IPV4 address: 300.0.0.0/8
...
ERROR: One or more invalid IP addresses were found ...   # exit 1
```

The fix strips the CIDR suffix before validating octets in `valid_ip()`, and applies the same 0-255 check in `valid_cidr()`. Valid aligned CIDRs, unaligned-but-in-range CIDRs like `10.0.5.7/24`, and plain addresses all behave exactly as before. IPv6 handling is untouched.

## Type of Changes

- Bugfix (Infrastructure / Tools)

## Testing Instructions

Run the script directly with a single argument:

- Invalid octets in CIDR form now warn and exit non-zero: `bash scripts/validate-cidr "300.0.0.0/8"` and `bash scripts/validate-cidr "10.0.0.999/8"`.
- Valid entries are unchanged (exit 0, no warning): `bash scripts/validate-cidr "10.0.0.0/8"`, `bash scripts/validate-cidr "192.168.1.0/24"`, `bash scripts/validate-cidr "1.2.3.4"`.

No Jira ticket, this came out of an external code read. Happy to adjust anything to fit your conventions.


akf (commented):
Thanks @arpitjain099. This looks good to me. We like to have two team members approve (usually one is the requester and the other the approver), so I'm commenting and will have a second person sign off.

arpage (approved):
Successfully tested w/ our IP address lists as well as the addresses provided by @arpitjain099 

Thank you @arpitjain099 - your contributions are much appreciated

PR #2837, opened 2026-08-07
USAGOV-2769 Split off any chained functions occurring after a .click() into their own calls

arpage:
## Jira Task

https://gsa-standard.atlassian-us-gov-mod.net/browse/USAGOV-2769

## Description

Split off any chained functions occurring after a .click() into their own calls to avoid asynchronous operation issues.

This will hopefully prevent timeouts occurring during tests using such chains.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [x] Other

## Testing Instructions

1. Go to https://github.com/usagov/usagov-2021/actions
2. Run Prod A11Y Regression Tests.  Instead of dev, use `USAGOV-2769-cypress-click-chaining`
3. Run Stage A11Y Regression Tests.  Instead of dev, use `USAGOV-2769-cypress-click-chaining`

Each environment's test should pass several runs - run each env 2 or 3 times.  I typically alternate between stage and prod.

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

6. Find the commit of this pull request.
7. Build and deploy the changes.
8. Update the Jira ticket by changing the ticket status to `Review in Test` and add a comment. State whether the change is already visible on [cms-dev.usa.gov](http://cms-dev.usa.gov/) and [beta-dev.usa.gov](http://beta-dev.usa.gov/), or if the deployment is still in process.

### DIFF
--- a/automated_tests/e2e-cypress/cypress/e2e/regression_testing/state_directory.cy.js
+++ b/automated_tests/e2e-cypress/cypress/e2e/regression_testing/state_directory.cy.js
@@ -17,11 +17,8 @@ paths.forEach((path, idx) => {
       });
       it("BTE 50A: Landing page: state drop-down", () => {
         // testing dropdown menu on click
-        cy.get("[data-test='stateInput']")
-          .click()
-          .then(($p) => {
-            cy.get("[data-test='stateDropDown']").should("be.visible");
-          });
+        cy.get("[data-test='stateInput']").click();
+        cy.get("[data-test='stateDropDown']").should("be.visible");
       });
 
       it("BTE 50B: Landing page: state drop-down", () => {
@@ -77,11 +74,8 @@ context("50 State Pages", () => {
 
   it("BTE 52A: 50-state pages: state drop-down", () => {
     // testing dropdown menu on click
-    cy.get("[data-test='stateInput']")
-      .click()
-      .then(($p) => {
-        cy.get("[data-test='stateDropDown']").should("be.visible");
-      });
+    cy.get("[data-test='stateInput']").click();
+    cy.get("[data-test='stateDropDown']").should("be.visible");
   });
 
   it("BTE 52B: 50-state pages: state drop-down", () => {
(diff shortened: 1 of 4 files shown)


automated_tests/e2e-cypress/cypress/e2e/regression_testing/state_directory.cy.js
```diff
@@ -17,11 +17,8 @@ paths.forEach((path, idx) => {
       });
       it("BTE 50A: Landing page: state drop-down", () => {
         // testing dropdown menu on click
-        cy.get("[data-test='stateInput']")
-          .click()
-          .then(($p) => {
```
akf, line 22: Seeing this gives me a different idea! The `.then()` should actually be helping here. It defines a callback, which should force the test to wait until `click()` has completed to execute. 

https://docs.cypress.io/api/commands/then

Conversely, I don't think separating chained functions into separate calls makes a difference, but you might have some knowledge I don't. https://docs.cypress.io/api/commands/click specifically says "It is [unsafe](https://docs.cypress.io/app/core-concepts/retry-ability#Only-queries-are-retried) to chain further commands that rely on the subject after .click()." Still, I'm unsure whether that applies! 

I think we should try the `then()` pattern more generally, but this also looks worth applying -- on the assumption that we're currently getting the URL before the location has updated: 

https://docs.cypress.io/app/core-concepts/retry-ability#Correctly-waiting-for-values

akf (commented):
I see that you tried running the tests using this branch and they failed. Looking at these changes gave me another idea though; see my comment on state_directory.cy.js

akf (approved):
Thanks! I can see this tests the same things, and that the tests ran cleanly.

PR #2838, opened 2026-08-07
USAGOV-2834: Load the OWASP CRS in the WAF (detection-only)

arpage:
## Why

Two CSV exports of a week of ModSecurity output (2026-07-20 → 26, prod/stage/dev) were reviewed to decide whether they represented real attacks and therefore deserved remediation effort.

**They do not.** They are malformed-content warnings. But the triage turned up something that does: **the WAF has been running 7 rules instead of the OWASP CRS.**

## The log triage (the original question)

Only two rules account for every alert, both from the CRS *recommended base configuration* — hygiene rules, not attack signatures:

| Rule | Count | What it is |
|---|---|---|
| `200002` REQBODY_ERROR | 148 | "could not parse the request body" → fail closed, 400 |
| `200003` MULTIPART_STRICT_ERROR | 58 | multipart body failed strict RFC validation → 400 |
| *(startup notice)* | 12 | WAF process start |

Three independent reasons to call this benign:

1. **No attack signatures fired at all** — zero `9xxxxx` CRS rules, because none were loaded.
2. **The multipart flags rule out evasion.** 53 of 58 hits are `LF 1` and nothing else — bare LF instead of CRLF line endings, i.e. a non-conforming HTTP client. The impedance-mismatch flags that indicate deliberate parser-confusion attacks (`DB`, `DA`, `BQ`, `BW`, `SM`, `IQ`, `IH`) are **zero in all 218 records**.
3. **Volume and sources are background noise.** ~30/day across three environments. 72 of the 76 XML errors are a single synthetic probe firing hourly at HH:23 with an XML content-type and no XML body. The rest is `*.app.cloud.gov` enumeration — one IP hits prod, stage and dev within the same second.

`severity:2` on `200002` is a hardcoded literal in the stock config, not a computed threat score.

> Note: the WAF is a cloud.gov route service, so the `GET /` / `POST /` in the request-line field reflects the forwarding mechanism — the real URL rides in `X-CF-Forwarded-Url`. That field was not used to reach any conclusion above.

## The actual finding

`Dockerfile-waf` downloads CRS 4.10.0 and `activate-rules.sh` tunes it, but nothing ever included it. The runtime log said so plainly: `rules loaded inline/local/remote: 0/7/0`.

Broken two independent ways:

- `owasp-modsecurity-main.conf` — the file nginx actually loads — included only `modsecurity.conf`, a 100%-commented override, and a **live `testparam` test rule** that returned 403 for any request carrying `?testparam=test`.
- `setup.conf` *does* have the right includes, but is referenced by nothing **and** points at `/etc/modsecurity.d/owasp-crs/`, a path the image never creates.

Also surfaced: the vendored `REQUEST-900` exclusions contain real USAGov tuning (`tx.crs_exclusions_drupal=1`, a wp-login exclusion) that **has never once executed**.

## What changed

| File | Change |
|---|---|
| `owasp-modsecurity-main.conf` | Include `crs-setup.conf` + `rules/*.conf`, ordered so our pre-CRS exclusions load first; drop the `testparam` rule; force `949110`/`959100` to `pass` |
| `modsecurity-override.conf` | Audit log → stdout, JSON, `RelevantOnly`, parts `ABHZ` |
| `setup.conf` | Fix the dead include paths; document that it is not the loaded file |
| `activate-rules.sh` | Document the silently no-op seds (see below) |
| `Dockerfile-waf` | Body limit `131072` → `320000` to match the hardcoded value; note the `MODSEC_*` vars are inert |
| `mod-sec-rules.conf` | Deleted — unreferenced; its `SecRuleRemoveById 949110` is superseded |

### Why audit logging had to be enabled

This is the non-obvious part, and it invalidated the original rollout plan.

**With `SecAuditEngine Off`, ModSecurity-nginx writes to the nginx error log only when it actually blocks a request.** A detection-only rollout is therefore completely silent. Verified locally — all three non-blocking variants detected a SQLi probe (anomaly score 8 vs threshold 5) and reported **nothing**:

| Config | Blocked? | Visible? |
|---|---|---|
| `SecRuleRemoveById 949110` | no | **no** |
| `SecRuleUpdateActionById 949110 "pass"` | no | **no** — `log` wrote to the disabled audit log |
| `SecRuleEngine DetectionOnly` | no | **no** — and it also stopped `200002` blocking (400 → 502) |

Enabling the audit log to stdout is what makes this stage worth deploying. Parts `C` and `E` are excluded so request bodies are never captured wholesale.

### Why blocking behavior is unchanged

`949110`/`959100` are the CRS blocking-evaluation rules that turn an anomaly score into a 403. Forcing them to `pass` leaves every CRS rule matching, scoring and logging while blocking nothing — and leaves `200002`/`200003` blocking exactly as they do in production today. Global `SecRuleEngine DetectionOnly` would additionally have stopped those from blocking, silently changing current behavior.

## Verification

Built and run locally from this branch (`docker build -f .docker/Dockerfile-waf .docker`):

| Check | Result |
|---|---|
| Rules loaded | **`0/807/0`** (was `0/7/0`) |
| SQLi / XSS / traversal / RCE | detected — `942100`, `941100/110/160/390`, `930100/110/120`, `932160` |
| Anomaly scoring | `949110` fired at scores 5, 20, 30 |
| Blocking | none — all probes returned upstream status, not 403 |
| `200002` malformed body | still **400**, unchanged |
| `?testparam=test` | no longer 403 |
| Request bodies in audit output | **0 occurrences** |

## Next steps

1. **Deploy to dev first and expect false positives.** This is the point of the detection-only stage. Tune via `REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf`.
2. **Watch audit-log volume on New Relic.** `RelevantOnly` with no status filter means only rule-triggered transactions are logged, but steady scanning traffic will produce some flow. Unmeasured so far.
3. **Exercise CMS login, benefit-finder submission, and site search** on dev/stage before promoting — the Drupal exclusions go live for the first time and are untested against real traffic.
4. **Then flip to blocking**, dev → stage → prod, by commenting out the two `SecRuleUpdateActionById` lines one environment at a time.

### Deliberately not done

- **`PARANOIA=2` is not applied.** Several seds in `activate-rules.sh` no longer match CRS 4.10.0 and silently no-op (`sed` exits 0 either way), so CRS runs at PL1 from the `REQUEST-901-INITIALIZATION` defaults. PL1 is the right level for a first rollout, so this is documented in-file rather than "fixed" into a much noisier PL2. `ANOMALY_INBOUND`/`ANOMALY_OUTBOUND` are equally inert.
- **`Dockerfile-waf-alpine` would now fail to start** — it copies these snippets but never downloads the CRS, so the literal `Include` is fatal. It is built by nothing (`container-build` only uses `Dockerfile-waf`) and referenced nowhere. Left alone rather than modify an untestable image; worth deleting if it is genuinely dead.
- The two CSV exports were left untracked rather than committed.

### Privacy note

Excluding part `C` prevents wholesale request-body capture, but a matched rule's `logdata` still quotes the offending fragment (e.g. `Matched Data: ... found within ARGS:id: 1' OR '1'='1`). That is the evidence the alert exists to convey, but it does mean a matched argument value can appear in logs. Part `B` also carries cookies. Both are documented in `modsecurity-override.conf`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


akf:
Wow. The **0/7/0** is a definite red flag.  I know that some of this used to work: 

> Also surfaced: the vendored REQUEST-900 exclusions contain real USAGov tuning (tx.crs_exclusions_drupal=1, a wp-login exclusion) that has never once executed.

I know I used to see those logged back in the day.

.docker/Dockerfile-waf
```diff
@@ -26,7 +33,7 @@ ENV APP_NAME=USAGov \
     MODSEC_PCRE_MATCH_LIMIT_RECURSION=500000 \
     MODSEC_REQ_BODY_ACCESS=on \
     MODSEC_REQ_BODY_LIMIT=13107200 \
-    MODSEC_REQ_BODY_NOFILES_LIMIT=131072 \
+    MODSEC_REQ_BODY_NOFILES_LIMIT=320000 \
```
akf, line 36: Hmmm. I'm in favor of removing these values if they're not used.

.docker/src-waf/opt/owasp-crs/modsecurity-override.conf
```diff
@@ -1,11 +1,33 @@
...
+# auditlog, so detections are captured even though they return 200. Ordinary
+# traffic produces no audit records, which keeps ingest volume down and avoids
+# duplicating the 4xx/5xx already covered by the access log.
+SecAuditEngine RelevantOnly
+SecAuditLog /dev/stdout
+SecAuditLogFormat JSON
+SecAuditLogType Serial
+# Parts: A=header, B=request headers, H=trailer (contains the rule Messages), Z=terminator.
+# Deliberately EXCLUDES part C (request body) and E (response body) so that user
+# form input -- CMS logins, benefit-finder answers -- is not captured wholesale.
```
akf: Just by-the-by, I don't think benefit-finder answers are ever submitted to the server; it's all client-side. 

However, some of the CMS request bodies are HUGE. For example, when a content editor reorganizes the menus and submits the result, a very large POST request is made. We would not want to log that.

.docker/src-waf/opt/modsecurity/activate-rules.sh
akf, line 1: So you're leaving them here with the intent to correct them later? (I read the PR description; I just forget stuff.)

.docker/src-waf/etc/nginx/snippets/owasp-modsecurity-main.conf
```diff
@@ -1,5 +1,45 @@
-# Include the recommended configuration
+# This is the file nginx actually loads; see conf.d/default.conf
```
akf, line 1: "This is the file nginx actually loads" is a bit weird here, without the context of the file nginx doesn't load. 
I do like the rest of the comment, though.

akf (commented):
Commenting only for now. 

This looks really promising! It seems likely that we lost the rules when we switched over from using the pre-build modsecurity image (because the modsecurity team deprecated it). I think you alluded to that. 

It looks like you deployed this to `dr` and I'm sorry to say I see this when I restart the waf in `dr` and then immediately check logs: 

```
% cf logs waf --recent | grep -i modsecurity
   2026-08-14T17:10:44.05-0700 [APP/PROC/WEB/0] OUT SFTWR_AUDIT: app=waf space=dr os="Debian GNU/Linux 13 (trixie)" nginx=1.30.1 modsecurity=3.0.13 crs=4.10.0 zlib=1.3.2 libxml2=2.13.6
   2026-08-14T17:10:44.07-0700 [APP/PROC/WEB/0] ERR 2026/08/15 00:10:44 [notice] 7#7: ModSecurity-nginx v1.0.3 (rules loaded inline/local/remote: 0/7/0)
``` 

So that might be a problem! But I'm in favor of deploying this branch to dev before merging it, to see how it goes. 

I'm a little leery of SecAuditLog, but your approach seems to have accounted for excessive logging. I tried to enable that some time ago, and perhaps that was after we'd broken the whole path for getting our customizations into this. At the time it either didn't work at all, or I couldn't get it to log to STDOUT. I forget which. Obviously if it winds up logging to the local filesystem that would be a problem.

arpage:
@akf 

I'm attaching links to:
1. 24 hours worth of [ModSec logs from Dev,](https://docs.google.com/spreadsheets/d/1_jLQcDxQyMzXMUOGChaLSWAoZpgD9IMnyc658EpB0tw/edit?gid=973206076#gid=973206076) with the new rules in place
2. A short section of [ModSec logs from Dev](https://docs.google.com/spreadsheets/d/16hOsPFYK0ZxiV6ds3b6ublhkr5ZCKS7H4fkB4FypiiU/edit?gid=1594867764#gid=1594867764), taken from the period when I was editing content. I edited 2 benefit finder entities (form and page - both appeared to process lots of json), and I edited a basic page.

It seems like we are not seeing an unreasonable number of log entries, however I think in prod we'll go over the 5k cvs download limit for a day - so we'll want to have some alternative way of gathering logs for auditing.

.docker/src-waf/opt/owasp-crs/modsecurity-override.conf
arpage: @akf - it looks like these are not logged. See attached csv logs.

.docker/src-waf/etc/nginx/snippets/owasp-modsecurity-main.conf
arpage: :) Oh! Claude - Why Art  Thou?

akf (approved):
So much learning here! Thank you.

PR #2846, opened 2026-08-24
USAGOV-2846-dependency-update-scripts: dependency checking and updati…

jacobaaronyeager:
…ng scripts

## Jira Task

https://gsa-standard.atlassian-us-gov-mod.net/browse/USAGOV-2846

## Description
Creates scripts for checking dependency updates and actually doing them, particularly for the weekly audit.

## Type of Changes

- [ ] New Feature
- [ ] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

- Test instruction 1
- Test instruction 2
- Test instruction 3

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality


bin/audit/deps-check
akf, line 1: I had a problem with this one. When I run it with anything but  `--composer-only` I get a couple of error messages, and my report stops after the table heading in the "Summary" section. Here's what I ran and the error messages I saw: 

```
~/dev/usagov-2021 %  bin/audit/deps-check --npm-only > deps-npm.md
»» npm outdated — Drupal theme...
»» npm audit — Drupal theme...
»» npm outdated — Benefit Finder...
»» npm audit — Benefit Finder...
»» npm outdated — Analytics Reporter...
»» npm audit — Analytics Reporter...
jq: error: syntax error, unexpected label, expecting IDENT or __loc__ (Unix shell quoting issues?) at <top-level>, line 19:
if has("__error__") then "| npm — \($label) | ? | ? | ? | ? | ? |"                                       
jq: error: Possibly unterminated 'if' statement at <top-level>, line 19:
if has("__error__") then "| npm — \($label) | ? | ? | ? | ? | ? |"
jq: 2 compile errors
```

I poked around a little but I don't think I have a handle on what should be happening here.

bin/audit/deps-update
akf, line 1: Since this relies on `deps-check`, it also fails with the above error. I tried running it with the `--composer-only` flag, but it still ran `npm audit` and failed at the summarize step. (It appears that the `--composer-only` flag applies to the interview/update part of this script, but not to calling `deps-check`.

akf (changes requested):
These failed for me -- see comments. It might be something about my system. If you don't see the same problem, let's go over it together.

bin/audit/deps-check
jacobaaronyeager: It was largely jq compatibility again. `label` is a reserved word in jq 1.6, so the variable `$label` fails to
compile there (jq 1.7+ accepts it). In `deps-check` that broke the Summary table on every run except --composer-only, truncating the report. In `deps-update` the same bug sat inside a `2>/dev/null || true` and silently dropped every npm vulnerability ticket from the final summary. So I renamed the jq args to `$area` / `$scope`, with a comment so they stay renamed.

`deps-update` also did not pass its scope on to `deps-check`, so `--composer-only` still paid for, and then failed on, the npm half of the collection. Passing it through exposed unguarded plan generation, which read raw files the skipped half no longer writes; the skipped half now gets an empty plan, so downstream readers need no scope guards of their own. A `--raw-in` dir holding only half the data now says so plainly instead of failing inside jq.

akf: Thanks for jumping on that so quickly!

akf (approved):
These are pretty slick!

PR #2863, opened 2026-09-04
USAGOV-2786: preventing "empty heading" agency-index pages

akf:
## Jira Task

https://gsa-standard.atlassian-us-gov-mod.net/browse/USAGOV-2786

## Description
This is a fix for a bug I haven't been able to  reproduce. 

### The problem: 
Sometimes Agency index pages wind up on the static site with empty boxes where the letter heading (an h2) and the "next" and "previous" letters should be. There are some screenshots attached to the Jira ticket. This is intermittent. 

### My theory about why this happens: 
I think the rendering of two or more pages can overlap when Tome is generating the static site pages. I think this mainly because it would explain the problem. I am fuzzy on: 
 - whether pages might be rendered in multiple threads in the same process
 - whether page rendering might be interleaved in some other way 

Agency index pages use two "view" blocks. One contains all the letters that are present as pages in the view. The other contains the entries for agencies that should be displayed on this page. The letter heading and next/previous navigation  are rendered with the second block, but the information to calculate next and previous is only present in the first block. 

So, we have a preprocess hook for each of these blocks. The first figures out what the current, next, and previous letters are and stashes them in a `drupal_static` variable; the second accesses that variable. Crucially, the first sets that variable to an empty array before populating it. If rendering of pages overlaps, it would be possible for a page to pick up that variable while it is an empty array. 

### How this PR addresses the problem

1. Minor modifications to `usa_twig_vars.module` add the "current letter" and language code to the string by which the static variable is accessed. This will prevent overlapping requests (for different lettered pages) from accessing each other's data. 
2. Adds a clause to `tome-sync.sh` that greps for the pattern of an empty heading in an agency index page and emits a warning if that pattern is found. I plan to set up an alert that will trigger on this warning so I can investigate further when it happens. 

Since I haven't figured out how to reproduce this problem, I modified the `tome-sync.sh` change locally to warn on any letter heading, so I could show what that output would look like. The attached `example-warnings.txt` shows that. 

[example-warnings.txt](https://github.com/user-attachments/files/31815154/example-warnings.txt)

## Type of Changes

- [ ] New Feature
- [x] Bugfix
- [ ] Documentation
- [ ] Infrastructure
  - [ ] CMS
  - [ ] WAF
  - [ ] WWW
  - [ ] Egress
  - [ ] Tools
  - [ ] Cron
- [ ] Other

### Change Requirements

- [ ] Requires New Config
- [ ] Requires New Content

### Validation Steps

All I know how to reliably test is that the agency indexes (both English and Spanish) generate correctly. I would recommend deploying this branch to dev and letting static site generation run, then reviewing the result. 

## Security Review

- [ ] Requires SSPP updates

## Reviewer Reminders

- Reviewed code changes
- Reviewed functionality

4. Find the commit of this pull request.
5. Build and deploy the changes.
6. Update the Jira ticket by changing the ticket status to `Review in Test` and add a comment. State whether the change is already visible on [cms-dev.usa.gov](http://cms-dev.usa.gov/) and [beta-dev.usa.gov](http://beta-dev.usa.gov/), or if the deployment is still in process.

### DIFF
--- a/scripts/tome-sync.sh
+++ b/scripts/tome-sync.sh
@@ -417,14 +417,16 @@ fi
 # Detect and warn only.
 #
 
-EMPTY_HEADING_LIST=""
-EMPTY_HEADING_PROBLEM=0
+EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep -HF '<h2 class="usagov-directory-letter-heading" tabindex="-1" id=""></h2>')
 
-EMPTY_HEADING_COUNT=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>' | wc -l)
+OLD_IFS=$IFS
+IFS=
+EMPTY_HEADING_COUNT=$(echo $EMPTY_HEADING_LIST | wc -l)
+IFS=$OLD_IFS
 
 if [ $EMPTY_HEADING_COUNT -gt 0 ]; then
   echo "WARNING: *** Empty headings found in agency index, total $EMPTY_HEADING_COUNT. List follows ***" | tee -a $TOMELOG
-  EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>')
+  # This will be one long line; that's okay.
   echo "$EMPTY_HEADING_LIST" | tee -a $TOMELOG
 fi
 
--- a/scripts/tome-sync.sh
+++ b/scripts/tome-sync.sh
@@ -413,20 +413,18 @@ fi
 ##############################################################
 
 ##############################################################
-# Empty headings in agency indes (implies also empty nav links), USAGOV-2786
-# Detect and warn only, at present!
+# Empty headings in agency index (implies also empty nav links), USAGOV-2786
+# Detect and warn only.
 #
 
 EMPTY_HEADING_LIST=""
 EMPTY_HEADING_PROBLEM=0
 
-# EMPTY_HEADING_COUNT=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>' | wc -l)
-EMPTY_HEADING_COUNT=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading"' | wc -l)
+EMPTY_HEADING_COUNT=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>' | wc -l)
 
 if [ $EMPTY_HEADING_COUNT -gt 0 ]; then
   echo "WARNING: *** Empty headings found in agency index, total $EMPTY_HEADING_COUNT. List follows ***" | tee -a $TOMELOG
-  #  EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>')
-  EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading"')
+  EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>')
   echo "$EMPTY_HEADING_LIST" | tee -a $TOMELOG
 fi
 
--- a/scripts/tome-sync.sh
+++ b/scripts/tome-sync.sh
@@ -412,7 +412,27 @@ fi
 # End of blog page checks (USAGOV-2667)
 ##############################################################
 
+##############################################################
+# Empty headings in agency indes (implies also empty nav links), USAGOV-2786
+# Detect and warn only, at present!
+#
 
+EMPTY_HEADING_LIST=""
+EMPTY_HEADING_PROBLEM=0
+
+# EMPTY_HEADING_COUNT=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>' | wc -l)
+EMPTY_HEADING_COUNT=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading"' | wc -l)
+
+if [ $EMPTY_HEADING_COUNT -gt 0 ]; then
+  echo "WARNING: *** Empty headings found in agency index, total $EMPTY_HEADING_COUNT. List follows ***" | tee -a $TOMELOG
+  #  EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>')
+  EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading"')
+  echo "$EMPTY_HEADING_LIST" | tee -a $TOMELOG
+fi
+
+#
+# End of Empty headings checks
+##############################################################
 
 if [ "$TOME_PUSH_NEW_CONTENT" == "1" ]; then
   echo "Pushing Content to S3: $RENDER_DIR -> $BUCKET_NAME/web/" | tee -a $TOMELOG
--- a/web/modules/custom/usa_twig_vars/usa_twig_vars.module
+++ b/web/modules/custom/usa_twig_vars/usa_twig_vars.module
@@ -228,13 +228,10 @@ function usa_twig_vars_preprocess_views_view_summary__federal_agencies(array &$v
   $current_letter = $view->getRequest()->query->get('letter') ?: 'a';
   $current_letter = mb_strtoupper($current_letter);
 
-  // We use $current_letter as part of the static key to avoid contention with
-  // other overlapping requests; see USAGOV-2786.
   $static_stash_name = usa_twig_vars_get_federal_agencies_static_name($current_letter);
   drupal_static_reset($static_stash_name);
   $shared = &drupal_static($static_stash_name);
   $shared = [];
-
   $shared['current'] = $current_letter;
 
   // Figure out next and previous letters.
@@ -256,15 +253,13 @@ function usa_twig_vars_preprocess_views_view_summary__federal_agencies(array &$v
 /**
  * Add context to the federal_agencies view.
  *
- * Allow us to show the current letter, and add links to the "next" and
+ * Enables us to show the current letter, and add links to the "next" and
  * "previous" letters.
  *
  * @param array<string, mixed> $variables
  * @param array<string, mixed> $info
  */
 function usa_twig_vars_preprocess_views_view_list__federal_agencies(array &$variables, string $hook, array $info): void {
-  // We use the requested "current_letter" as part of the static key to avoid contention with
-  // other overlapping requests; see USAGOV-2786.
   $current_letter = $variables['view']->getRequest()->query->get('letter') ?: 'a';
   $current_letter = mb_strtoupper($current_letter);
   $static_stash_name = usa_twig_vars_get_federal_agencies_static_name($current_letter);
@@ -275,9 +270,9 @@ function usa_twig_vars_preprocess_views_view_list__federal_agencies(array &$vari
 /**
  * Get a consistent name for the static variable we use to share
  * letter context between two preprocess hooks for the federal
- * agencies view. To avoid collisions between page renders,
- * we include current displayed letter and the site language,
- * in addition to a unique string based on the function that
+ * agencies view. To avoid collisions between page renders (see
+ * USAGOV-2786), we include current displayed letter and the site
+ * language, in addition to a unique string based on the function that
  * populates this variable.
  *
  * @param string $letter
--- a/web/modules/custom/usa_twig_vars/usa_twig_vars.module
+++ b/web/modules/custom/usa_twig_vars/usa_twig_vars.module
@@ -230,9 +230,9 @@ function usa_twig_vars_preprocess_views_view_summary__federal_agencies(array &$v
 
   // We use $current_letter as part of the static key to avoid contention with
   // other overlapping requests; see USAGOV-2786.
-  $our_unique_name = implode("--", ['usa_twig_vars_preprocess_view_federal_agencies', $current_letter]);
-  drupal_static_reset($our_unique_name);
-  $shared = &drupal_static($our_unique_name);
+  $static_stash_name = usa_twig_vars_get_federal_agencies_static_name($current_letter);
+  drupal_static_reset($static_stash_name);
+  $shared = &drupal_static($static_stash_name);
   $shared = [];
 
   $shared['current'] = $current_letter;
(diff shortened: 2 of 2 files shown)


scripts/tome-sync.sh
```diff
@@ -412,7 +412,25 @@ fi
...
 
+##############################################################
+# Empty headings in agency index (implies also empty nav links), USAGOV-2786
+# Detect and warn only.
+#
+
+EMPTY_HEADING_LIST=""
+EMPTY_HEADING_PROBLEM=0
+
+EMPTY_HEADING_COUNT=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>' | wc -l)
```
jacobaaronyeager: This pattern can't match — the check will never fire.

An empty heading renders as `id=""`, not `id`. The template is
`id="{{- alpha_context.current -}}"`, and Twig renders a null / empty-array /
missing-key `alpha_context` to the empty string, leaving the `=""` in place:
```
    <h2 class="usagov-directory-letter-heading" tabindex="-1" id=""></h2>
```
Nothing downstream strips it — the only HTML rewrites in this script are the
URL `sed`s at lines 116 and 153-155.

Running this exact `find | xargs | grep` against a fixture containing the real
markup, inside the cms container:
```
    PR's pattern (id>)  -> 0 matches
    id="" pattern       -> 2 matches
```
Suggest matching the rendered form, and using `-F` since the pattern is a fixed
string:
```
    grep -HF '<h2 class="usagov-directory-letter-heading" tabindex="-1" id=""></h2>'
```
Worth adding a fixture to confirm the check fires before we rely on it.

scripts/tome-sync.sh
```diff
@@ -412,7 +412,25 @@ fi
...
 
+if [ $EMPTY_HEADING_COUNT -gt 0 ]; then
+  echo "WARNING: *** Empty headings found in agency index, total $EMPTY_HEADING_COUNT. List follows ***" | tee -a $TOMELOG
+  EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep '<h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>')
```
jacobaaronyeager: Two things on this second `find`:

1. It's a duplicate full-tree scan. Lines 423 and 427 walk every `index.html`
   under `/var/www/html` and grep them twice — 2,178 files on my local build,
   considerably more in prod. Capture the list once and derive the count from it.

2. Missing `-H`. If an xargs batch ends up holding exactly one file, busybox
   grep omits the filename prefix, so that entry's page path is lost from the
   list — which is the whole point of the list. Confirmed in the cms container.

Also, for awareness rather than a change request: both `find`s scan
`/var/www/html` rather than `$RENDER_DIR`, the snapshot actually pushed to S3.
That matches the existing `BLOG_DIR` check above, so it's consistent with
convention here — just noting the two can drift, since the script itself warns
that tome's output dir isn't locked.

web/modules/custom/usa_twig_vars/usa_twig_vars.module
```diff
@@ -252,17 +253,38 @@ function usa_twig_vars_preprocess_views_view_summary__federal_agencies(array &$v
...
+ * letter context between two preprocess hooks for the federal
+ * agencies view. To avoid collisions between page renders (see
+ * USAGOV-2786), we include current displayed letter and the site
+ * language, in addition to a unique string based on the function that
+ * populates this variable.
+ *
+ * @param string $letter
+ * @return string
+ */
+function usa_twig_vars_get_federal_agencies_static_name(string $letter): string {
```
jacobaaronyeager, line 281: The language component doesn't disambiguate the case the commit message
describes.

`getCurrentLanguage()` is the *current page* language, not the language of the
attachment being rendered. In `views.view.federal_agencies`, both attachments
attach to `page_1`:
```
    attachment_1 (Alpha List)    -> default, block_1, page_1
    attachment_2 (Alpha List ES) -> default, block_2, page_1
```
So on `page_1` the summary hook runs twice, both calls compute the *same* key
from the one page language, and the second overwrites the first — the exact
collision the language key was added to prevent.

In practice this isn't reachable: `/agency-index-view` returns 403 to anonymous
users, so it never lands in the static build, and the real pages (`block_1` /
`block_2`) each render a single attachment. So no change needed for this ticket
— but the rationale in the docblock doesn't hold for the one display where two
summaries actually collide, and it'd be worth saying so here so the next person
doesn't rely on it.

web/modules/custom/usa_twig_vars/usa_twig_vars.module
```diff
@@ -252,17 +253,38 @@ function usa_twig_vars_preprocess_views_view_summary__federal_agencies(array &$v
...
  *
  * @param array<string, mixed> $variables
  * @param array<string, mixed> $info
  */
 function usa_twig_vars_preprocess_views_view_list__federal_agencies(array &$variables, string $hook, array $info): void {
-  $alpha_context = &drupal_static('usa_twig_vars_preprocess_view_federal_agencies');
+  $current_letter = $variables['view']->getRequest()->query->get('letter') ?: 'a';
+  $current_letter = mb_strtoupper($current_letter);
+  $static_stash_name = usa_twig_vars_get_federal_agencies_static_name($current_letter);
+  $alpha_context = &drupal_static($static_stash_name);
```
jacobaaronyeager, line 266: What's the intended behavior when the summary hook hasn't run for this page?

`drupal_static($static_stash_name)` returns NULL for an unpopulated key, which
renders as the empty heading this ticket is about. Tome calls
`RequestPreparer::prepareForRequest()` — and therefore a bare
`drupal_static_reset()` — before every path, so the key starts unset on each
render and only the summary hook fills it. If the summary attachment is served
from the render cache while the list isn't, the letter-scoped key doesn't help.

Not saying that's what's happening in prod — I couldn't reproduce 2786 locally
(see my summary comment). Mostly asking whether a fallback here, deriving
`current` directly from the request rather than trusting the static, would make
this robust regardless of whether the summary hook ran.

jacobaaronyeager (changes requested):
The comments and commentary below are from Claude primarily.  My own examination of it's comments makes me think it's right, but YMMV.

~~~

Passing: `php -l`, phpcs (the one warning at line 308 is pre-existing, outside
the diff), phpstan, the live EN/ES agency index for letters a-b, and the
`federal_directory_2024` + `regression_testing/federal_directory` Cypress specs
(3/3, EN and ES).

I couldn't reproduce 2786 on local content. I built a harness mirroring
`StaticExportPathCommand::execute` — `tome_static.generator->requestPath()` with
`tome_static.request_preparer->prepareForRequest()` before each path — and ran
all 26 letters in both languages through a single process. 38 pages generated,
38 correct, 0 empty, 0 wrong, on this branch *and* on dev. So no regression, and
the 38/38 does confirm the two hooks computed matching keys on every EN and ES
page (a mismatch would have produced exactly the empty heading we're hunting).
But I can't confirm the fix changes anything on this dataset.

Given that, the tome-sync detector is doing the real work here — and as written
it can't fire (see my comment on line 423). I'd want that pattern fixed and
verified against a fixture before this merges, otherwise we get silence either
way and won't know whether the module change helped.

One aside in case it saves someone time: an early version of my harness that
omitted `prepareForRequest()` showed every page inheriting the first one's
letter. That was my harness bug, not the code — `drupal_static_reset()` inside
`prepareForRequest` clears it.

scripts/tome-sync.sh
akf: Thanks, this is a good catch. I don't seem to have saved actual HTML output when this happened (although I did see it once), so I was relying on what SiteImprove showed me. I'm pretty sure it's not showing the raw HTML. 

(There is actually potential rewriting in a Tome listener. Depending on the page content, at least one of the listeners will load a page into a DOM document, modify the DOM, and then write out HTML from that, which does sometimes make subtle changes. I don't think that applies here.)

web/modules/custom/usa_twig_vars/usa_twig_vars.module
akf: If the summary hook hasn't run, we're out of luck. As far as I could tell the "next" and "previous" letters weren't available from the information present at this point (we can get "current," but things are still broken). Maybe there's a way to reach around through the view and derive them.

You've prompted me to check whether we would get a different error in that case, though -- just by commenting out the line that assigns `$shared` to `&drupal_static($static_stash_name)`. I thought we would get an error because the twig file blithely assumes `alpha_context` will be an array, but it does in fact look like what happens when `alpha_context` is an empty array.

akf:
@jacobaaronyeager  

> One aside in case it saves someone time: an early version of my harness that
omitted `prepareForRequest()` showed every page inheriting the first one's
letter. That was my harness bug, not the code — `drupal_static_reset()` inside
`prepareForRequest` clears it.

This is really interesting, because the same function that populates the letters into the static variable calls `drupal_static_reset($static_stash_name)` first. It sounds as if omitting `prepareForRequest()` must bypass `usa_twig_vars_preprocess_views_view_summary__federal_agencies` on all but the first rendering of the view.

web/modules/custom/usa_twig_vars/usa_twig_vars.module
akf: I was unable to retrieve the language in any other way within that hook, so I decided to disambiguate this by detaching the Alpha List blocks from the `/agency-index-view` page. I would remove that page altogether, but it seems to be important to have something to provide a route for that view. 

I think that Page was my initial attempt to implement what we have now at `/agency-index` (and didn't yet have any Spanish-language content for the view). After fighting the existing structure to try to match the layout of non-view pages, I realized that just putting blocks on a "regular" page would work more smoothly.

scripts/tome-sync.sh
akf: I've removed the duplicate scan (honestly wasn't too concerned about it since it only occurs in a condition that shouldn't happen) and added the `-H` flag.

akf:
@jacobaaronyeager please give this another review. I held off because I wanted to think about "What's the intended behavior when the summary hook hasn't run for this page?" and whether that's a thing that can happen (on the pages we generate for anonymous users), and at this point I've thought and although I can't prove that can't happen, I would like to get the logging out there to monitor this.

scripts/tome-sync.sh
```diff
@@ -412,7 +412,27 @@ fi
...
+##############################################################
+# Empty headings in agency index (implies also empty nav links), USAGOV-2786
+# Detect and warn only.
+#
+
+EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep -HF '<h2 class="usagov-directory-letter-heading" tabindex="-1" id=""></h2>')
+
+OLD_IFS=$IFS
+IFS=
+EMPTY_HEADING_COUNT=$(echo $EMPTY_HEADING_LIST | wc -l)
```
jacobaaronyeager, line 424: `echo $EMPTY_HEADING_LIST | wc -l` returns 1 when the list is empty — echo of an empty variable still emits one newline. So `[ $EMPTY_HEADING_COUNT -gt 0 ]` is always true. Ran your block verbatim against a tree with zero empty headings:
```
WARNING: *** Empty headings found in agency index, total        1. List follows ***
```
The "List follows" being the important point here, as it demonstrates it was trying to print an empty tree when it was supposed to skip it. Every tome-sync run, forever. For N≥1 the count is right; it's wrong in exactly the case that matters. The `IFS=` dance isn't doing anything here either — `echo "$LIST" | wc -l` is equivalent.

scripts/tome-sync.sh
```diff
@@ -412,7 +412,27 @@ fi
 # End of blog page checks (USAGOV-2667)
 ##############################################################
 
+##############################################################
+# Empty headings in agency index (implies also empty nav links), USAGOV-2786
+# Detect and warn only.
+#
+
+EMPTY_HEADING_LIST=$(find /var/www/html -name index.html -print0 | xargs -0 grep -HF '<h2 class="usagov-directory-letter-heading" tabindex="-1" id=""></h2>')
```
jacobaaronyeager, line 420: This corrects what I told you last time. I said the markup is `id=""` so use `-F`. That's only true for pages that pass through untouched. TomeEventSubscriber.php on lines 203-213 re-serializes the whole document through `Masterminds\HTML5::saveHTML()` whenever `$changes` is `TRUE` — any link to exactly `/es`, `/es?…` or `/es#…`, or any `src` containing a space/`%20`. That serializer writes an empty attribute bare. Verified against the repo's own vendored Masterminds:
```
input : <h2 class="usagov-directory-letter-heading" tabindex="-1" id=""></h2>
output: <h2 class="usagov-directory-letter-heading" tabindex="-1" id></h2>
```
So your original `id>` instinct was right for re-serialized pages and my correction was right for the others — both forms occur, and `-F '…id=""></h2>'` catches only one. Whether a given agency-index page re-serializes depends on its link/image content, so this fails non-deterministically, which is the worst outcome for a monitor.

jacobaaronyeager, line 420: Line 58 snapshots `cp -Rp /var/www/html/* $RENDER_DIR`, and everything downstream — hostname rewrites, sitemap edit, the S3 push — works on `$RENDER_DIR`. Line 57's own comment flags that `/var/www/html` is unlocked and a concurrent Tome run can mutate it. So the check can report pages that were never published and miss pages that were. The script has already cd'd to `$RENDER_DIR`. (The blog check above has the same flaw — pre-existing, not yours.)

Tested replacement (silent on clean, reports both broken forms, silent when the dirs don't exist):
```
EMPTY_HEADING_LIST=$(find "$RENDER_DIR/agency-index" "$RENDER_DIR/es/indice-agencias" \
  -name index.html -exec grep -HE 'class="usagov-directory-letter-heading"[^>]*></h2>' {} + 2>/dev/null)

if [ -n "$EMPTY_HEADING_LIST" ]; then
  EMPTY_HEADING_COUNT=$(printf '%s\n' "$EMPTY_HEADING_LIST" | wc -l | tr -d ' ')
  echo "WARNING: *** Empty headings found in agency index, total $EMPTY_HEADING_COUNT. List follows ***" | tee -a $TOMELOG
  echo "$EMPTY_HEADING_LIST" | tee -a $TOMELOG
fi
```
Note: this block should address all three issues.

Keying on "the heading closes with no text" survives both serializations and any attribute reordering. `-n` on the string matches the idiom the blog check above already uses. Narrowing to the two directories drops the scan from 2,000+ files to ~50.

jacobaaronyeager (changes requested):
Claude has one other thing it would like you to look into:

Tome's concurrency is process-level — `StaticCommand::exportPaths()` spawns `drush tome:static-export-path` as separate `Symfony\Component\Process` children, each with its own PHP memory. Inside a child, `StaticExportPathCommand::execute()` calls `prepareForRequest()` before every path, and `RequestPreparer.php:102` is a bare `drupal_static_reset()` — it wipes all statics.

There's no window where one page's render reads another page's static: across processes nothing is shared, within a process the statics are gone before the next path starts. Scoping the key by letter+language is harmless (both hooks resolve the same key from the same `Request`, so it can't regress), but it can't be what fixes the empty headings. Which matters for your monitoring plan: if the warning stays quiet on dev, that won't tell you the bug is fixed.

What can leave `alpha_context` unset in a single request is the summary preprocess simply not running — the attachment didn't render. That's the question I raised last round, and this PR makes it concrete rather than hypothetical: detaching the attachments from `page_1/default` means `/agency-index-view` now renders exactly the markup the new grep hunts for. (Harmless in practice — `page_1` is role-restricted to editors/admins, so Tome, which requests as anonymous, never exports it. The export pipeline uses `attachment_1/attachment_2` via `execute()` directly, so it's unaffected by the `displays` change.)

The list hook now computes `$current_letter` itself, so the fallback is two lines and closes the visible defect regardless of cause:
```
$alpha_context = &drupal_static($static_stash_name);
if (empty($alpha_context['current'])) {
  $alpha_context['current'] = $current_letter;
}
$variables['alpha_context'] = $alpha_context;
```
You're right that `next/prev` aren't derivable from the list context — leaving those empty is fine. The heading is the reported bug.

scripts/tome-sync.sh
akf: > You're right that next/prev aren't derivable from the list context — leaving those empty is fine. The heading is the reported bug.

The missing next/prev are very important, and were reported on a separate ticket that I combined into this work. I'm sorry the title of the ticket gave that impression. (The missing heading also looked easier to detect.) 

I'm not sure how to verify that Tome's concurrency is process-level, but I think I might have some focus time for that soon. 

I've had some other ideas about how to research the problem and a different way to address it, so I'm going to close this PR and start over.

PR #2872, opened 2026-09-17
USAGOV-2511: Drupal 11 upgrade (phases 0-7 of 10)

arpage:
Work in progress toward Drupal 11. Phases 0-5 and 7 of a 10-phase plan; **do not merge yet**.

Target: `drupal/core-recommended ^11.4`, landing exactly **11.4.7** — the only 11.4 release that
satisfies the `drupal/s3fs` version bands, the Drupal 11.2 hook-ordering floor that
`node_menus dev-3.x` needs, and SA-CORE-2026-013 simultaneously.

## Landed

**Phase 1 — composer environment hygiene.** Resolution was running on PHP 8.5.9 without `ext-gd`
while production is 8.3.29 with it. `--ignore-platform-reqs` hid this, so the lock was being
resolved against an environment that does not exist. The real production platform is now declared
via `config.platform`, and the flag is gone from both `bin/composer` and `.docker/Dockerfile-cms`.

**Phase 2 — static analysis gate.** `usagov_benefit_finder` (215 files) was excluded from phpstan
and had never been analysed; that exclusion is removed and all 151 resulting errors are recorded
in `phpstan-baseline.neon`, so the suite passes and only *new* errors fail. Zero deprecation
findings in custom code. `phpunit.xml.dist` rewritten (it used a PHPUnit 4.1 schema and pointed at
a directory that does not exist).

**Phase 3 — every D10-capped contrib package resolved**, each verified independently on 10.6.15:

| package | change | note |
|---|---|---|
| ctools | 3.15.0 → 4.1.1 | major |
| field_defaults | 2.0.0 → 2.1.1 | |
| image_style_warmer | 1.2.0 → 1.3.0 | |
| log_stdout | 1.5.0 → 3.0.0 | major, no 2.x exists |
| new_relic_rpm | 2.1.1 → 2.3.0 | |
| diff | 1.8.0 → 2.1.1 | 🔒 SA-CONTRIB-2026-096 |
| paragraphs | 1.19.0 → 1.23.0 | 🔒 SA-CONTRIB-2026-060/061 |
| s3fs | 3.10.0 → 3.11.0 | required for the 11.4 band |
| address | **removed** | unused: absent from `core.extension.yml`, from `pm:list`, and zero field storages |
| content_lock | 2.4.0 → 3.0.0 | major + patch rewritten for 3.x |
| paragraphs_entity_embed | 3.0.1 → 4.0.0 | major |
| faqfield | 7.1.0 → 8.0.1 | major |

`composer audit`: **4 advisories → 1**. The remainder is SA-CORE-2026-013, fixed by 11.4.7 in phase 9.

**Phase 4 — custom extension metadata.** All 26 custom extensions (25 modules + the `usagov`
theme) now declare `core_version_requirement: ^10 || ^11`. Deliberately not `^11`: the site stays
on 10.6.15 until phase 9, so dropping `^10` would uninstall every custom extension immediately.

**Phase 5 — `ConfigFormBase` constructors.** The only removed-in-D11 API in ~11,700 lines of
custom PHP. `OrphanedEntitiesSettings` re-promoted `protected $typedConfigManager = NULL` over the
parent property, which against D11's typed parent is a **fatal at class-compile time** — the
module would have been unloadable. Dropping the `protected` keyword so the parent's property is
reused, not shadowed, is the actual fix; forwarding the argument alone would not have sufficed.
`LoginSettingsForm` simply lacked the parameter.

**Phase 7 — three real `settings.php` defects.** Scope reduced after checking 11.4.7 source
(see *Corrections* below). `state_cache` was set to an array, which is not a valid form — core
tests `!== TRUE`, so **state caching was off site-wide**, a DB query per state read on every
request, *and* it suppressed the warning designed to catch that. Also guarded an unchecked
`$_SERVER['argv'][1]`, and modernised the DB driver namespace so it no longer relies on the D8 BC
shim.

## Corrections to the original upgrade audit

Three assumptions were checked against real source and turned out wrong. Recording them so they
are not re-derived:

1. **The D8 database-driver BC shim still exists in 11.4.7** (`Database.php:203-221`). The legacy
   namespace would not have broken the upgrade. Demoted from blocker to hygiene.
2. **`update_free_access` is still present in 11.4.7** `default.settings.php:312`. Nothing to
   remove; dropped from scope.
3. **`BenefitFinderSettingsForm` needed no change.** The audit named three `ConfigFormBase`
   subclasses; only two were defective. It inherits core's `create()`, which already injects
   `config.typed` — proven empirically, as it was the only one of the three that never emitted the
   deprecation.

Consequently the planned wholesale rebase of `settings.php` onto D11's `default.settings.php` was
**deferred to phase 9**. Only ~150 of its 996 lines execute; rewriting a file that controls
database connectivity, trusted hosts and the redis bootstrap container to refresh stale comments
is risk without return — especially as trusted-host behaviour cannot be exercised locally at all.

## Verification

Per phase: `drush updb`, `drush cr`, `drush config:status` (drift exported and committed),
phpstan against the baseline, and HTTP 200 on `/` and `/user/login`.

Phase 5 was additionally verified by instantiating all three settings forms through the container
with an explicit `E_USER_DEPRECATED` handler — `@trigger_error` is otherwise silently suppressed,
so a naive probe reports a false all-clear. Before: two forms emit the `$typedConfigManager`
deprecation. After: all three silent, and reflecting `ConfigFormBase::$typedConfigManager` on each
instance yields a real `TypedConfigManager`, which is what proves the property is no longer
shadowed.

Phase 7: `state_cache` now reads `bool(true)`; the database connects as
`Drupal\mysql\Driver\Database\mysql\Connection` (the modern class, not via the shim) and a real
query returns 2231 nodes; zero-argument `drush` is warning-free, confirmed against an A/B control
proving the old form warns and that the container does surface such warnings.

Manual sandbox test plans for phases 4+5 and for phase 7, targeting the `dr` environment, are
posted as a comment on this PR.

## Not yet done

Phase 6 (rebase the two core patches and the coupled `views_menu_children_filter` patch), phase 8
(`node_menus` `dev-3.x` pin, which must be atomic with phase 9), phase 9 (the core bump), phase 10
(non-blocking modernization).

## Known gaps

- **No runnable test harness until phase 9.** `drupal/core-dev` at 10.6.15 requires
  `mglaman/phpstan-drupal ^1.2.12`, which would downgrade the PHPStan 2.x stack exactly when it is
  most useful; at 11.4.7 it accepts `^2.0.15`. phpstan is the only gate until then.
- **faqfield content preservation is unproven** — the local database has zero rows in
  `paragraph__field_faq`, so the 7.x → 8.x jump is verified for schema and config only.
- **The CKEditor embed dialog needs a human pass** (paragraphs_entity_embed 4.x, config-level
  verification only).
- **Rector is broken**, independent of this work: 46 custom files fail on a `phpstan/phpdoc-parser`
  conflict pulled in by `slevomat/coding-standard`. Diagnosed and documented in `rector.php`; left
  as its own ticket so it cannot destabilise the phpstan baseline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


arpage:
@arpage - note to self for next code session:

Everything from the pause checklist is now done. Both remaining items are yours when convenient: 
- move pre-phase3-backup.sql out of the repo root if a reclone is likely
- and pick up at Phase 4 using section D of the plan file.

arpage:
# Manual sandbox test plans — `dr`

Two plans: one for **phases 4 + 5** (custom extension metadata + ConfigFormBase), one for
**phase 7** (settings.php). Assumes Drupal admin familiarity, not familiarity with this upgrade.

## Environment

| | |
|---|---|
| Sandbox space | cloud.gov **`dr`** — it *is* in the `settings.php:824` whitelist, so it runs the real cloud.gov path: `cloud_split` on, `local_split` off, RDS TLS on, trusted hosts enforced |
| CMS | https://cms-dr.usa.gov |
| Public / static | https://beta-dr.usa.gov |
| Login | **SSO only.** `gsaauth.dr.conf` exists so SAML works. `/user/login` shows only a "Login via SSO" button unless State `usagov_login_local_form` is set |
| Deploy | **Manual.** CircleCI → `approve-build-and-push-container` → `approve-dr-deployment`. `dr` has no branch filter, so this branch can deploy there |
| Existing automation | GitHub Actions **"DR A11Y Regression Tests"** (`workflow_dispatch`) runs the Cypress `regression_testing/*.cy.js` suite against `beta-dr.usa.gov` |

⚠️ **If both plans are run against the same deploy, run Plan B first.** If the site is down from a
database-connection error, nothing in Plan A is meaningful.

Take a screenshot of `/admin/reports/status` **before** deploying — several steps compare against it.

---

## Plan A — phases 4 + 5

**Why it could break:** all 26 custom extensions now declare `core_version_requirement:
^10 || ^11`. A typo in any one makes that extension core-incompatible and Drupal **silently
refuses to load it** — no error, the feature just disappears. Separately, two settings-form
constructors changed signature; if wrong, those two admin pages throw.

| # | Step | Expected | Fail means |
|---|---|---|---|
| 1 | https://cms-dr.usa.gov → log in via SSO | CMS loads, authenticated | Deploy/SAML issue, not these phases |
| 2 | `/admin/modules`, filter `USA` | **24 custom modules present and enabled.** None says "incompatible with this version of Drupal" | ❌ a `core_version_requirement` typo |
| 3 | `/admin/appearance` | **USAgov** is the default theme, not marked incompatible | ❌ theme `.info.yml` typo |
| 4 | `/admin/reports/status` | Renders; no **new** warnings vs the pre-deploy screenshot | ⚠️ investigate |
| 5 | `/admin/config/content/orphaned-entities` | Loads: a fieldset per bundle, Clear / Delete / Do-nothing radios. *(Fieldset titles show machine names — known, not a bug)* | ❌ **phase 5 regression** |
| 6 | Change one radio → **Save** → reload | Saves without error; value persists | ❌ **phase 5 regression** |
| 7 | `/admin/config/system/login-settings` | Loads: SSO path, SSO text, "Display the local login form" | ❌ **phase 5 regression** |
| 8 | Edit the SSO link text → **Save** → reload | New value persists | ❌ **phase 5 regression** |
| 9 | Clear **SSO login path** → Save → `/admin/reports/status` | "USAgov Login Module" shows a warning about the SSO link | ❌ `hook_requirements` broken |
| 10 | **Restore the SSO path** → Save → recheck | Warning clears. **Do not skip — an empty SSO path can lock out login** | |
| 11 | `/admin/reports/dblog`, Error/Critical, since deploy | Nothing new naming `ConfigFormBase`, `TypedConfigManager`, or a `usa*` module | ❌ |
| 12 | On https://beta-dr.usa.gov spot-check: a basic page, the bilingual left nav, a wizard flow, the benefit finder | Renders as before | ❌ a module silently failed to load |

**Pass:** steps 2, 3, 5–8 clean and step 11 shows nothing new.
**Highest value: steps 2 and 6.** Step 2 is the only thing that catches a silently-dropped
extension; step 6 is the only thing that proves the constructor change round-trips.

---

## Plan B — phase 7

**Why it could break:** `settings.php` controls database connectivity, trusted hosts and caching.
The DB driver namespace change is the highest-blast-radius line in the repo — if wrong, **every
request fails**. Step 1 is effectively the whole test. Note the `state_cache` fix has **no visible
UI symptom either way**; it can only be confirmed from the CLI at step 5.

| # | Step | Expected | Fail means |
|---|---|---|---|
| 1 | Load https://cms-dr.usa.gov | **Site responds.** Not "Unable to connect to your database server", not a white screen | ❌ **STOP, roll back.** DB namespace/autoload wrong |
| 2 | Load https://beta-dr.usa.gov | Static site serves | ❌ |
| 3 | Check for a host error | No *"The provided host name is not valid for this server"* (HTTP 400) | ❌ trusted-host regression. **Only observable on cloud.gov — local cannot catch this** |
| 4 | `/admin/reports/status` | Renders; no new DB, redis or cache warnings | ⚠️ |
| 5 | `cf ssh` to the dr CMS app, then:<br>`drush php:eval "var_dump(\Drupal\Core\Site\Settings::get('state_cache'));"` | **`bool(true)`** — not an array, not NULL | ❌ fix did not land |
| 6 | Same shell: run `drush` with **no arguments** | No `Undefined array key 1`, no `str_starts_with(): Passing null` | ❌ argv guard wrong |
| 7 | Same shell:<br>`drush php:eval "var_dump(\Drupal\Core\Site\Settings::get('redis.connection')['scheme']);"` | `string(3) "tls"` on cloud.gov | ❌ VCAP parsing disturbed |
| 8 | Same shell:<br>`drush php:eval "print get_class(\Drupal\Core\Database\Database::getConnection());"` | `Drupal\mysql\Driver\Database\mysql\Connection` | ❌ namespace change wrong |
| 9 | Edit and save any node in the CMS | Saves; appears on the front end after the next static build | ❌ |
| 10 | `/static-site-status` | Normal state, not "Failed" | ❌ tome regression |
| 11 | Run **DR A11Y Regression Tests** in GitHub Actions | Passes at the same rate as a pre-deploy baseline run | ⚠️ compare, don't assume |
| 12 | `/admin/reports/dblog` since deploy | No new database, cache or settings errors | ❌ |

**Pass:** steps 1–3 clean, step 5 shows `bool(true)`, step 8 shows the modern connection class,
step 11 no worse than baseline.

**Rollback:** `git revert` the phase 7 commit and redeploy. `settings.php` is plain code with no
database or config dependency, so the revert is complete and immediate — unlike the contrib
phases, nothing needs `updb` or `cim` to undo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2877, opened 2026-09-22
Drop the /tmp debug log from the Tome image src rewrite

arpitjain099:
## Jira Task

None. I am an outside contributor and do not have access to the tracker, happy for this to be attached to a ticket if you need one.

## Description

Removes the two `file_put_contents('/tmp/tome-img-path-debug.log', ...)` calls from `TomeEventSubscriber::modifyHtml` along with the now-empty `else` that held the second one. The `src` rewrite itself is untouched.

Both arms of the USAGOV-2515 check log, including the arm for images that needed no change, and nothing guards them by environment, so the static build writes a line per element with a `src`, per page, on every environment rather than just locally.

## Type of Changes

- [x] Bugfix

## Testing Instructions

Run a Tome static build and confirm `/tmp/tome-img-path-debug.log` is not created. Image paths containing a space or `%20` should still come out with `+`.

For what it is worth, I ran the loop on its own over the current front page of www.usa.gov first: 15 elements with a `src`, 15 lines logged, every one of them the "No change src" line.


akf (commented):
Thanks for catching this! I approve. Per our policy, a second person on our team must also approve.

